PR #1, opened 2016-10-20
Add build support for maven and cmake

rleigh-codelibre:
Changes:

- Remove ant build files
- Move components/* to the top-level
- Each component uses a standard maven project layout
- Specification schemas and transforms are relocated under `src/main/resources`
- C++ code relocated under `src/main/cpp` and `src/test/cpp` so the C++ and Java code now sits side by side, to make the layout cleaner and more consistent; they can e.g. share resources and test data in the future
- Correct all paths to use the new locations
- xsd-fu module uses pom packaging to prevent it making a jar
- Top-level build is mostly derived from the ome-common-java pom, but with added use of `pluginManagement` for the sub-projects.

To do:

- Release process; we'll need to strip out genshi and the other python bits to make a `-dfsg` release as for bioformats.  The script needs copying over an adapting
- The C++ code is using git for versioning as for the other repos; we could update it to get the version from the toplevel pom to avoid inconsistency.

Testing:

- Build with maven
- Check out https://github.com/rleigh-codelibre/bioformats/tree/ome-model-split
- Build bioformats with ant and maven; it will use the specification and ome-xml jars

For C++:

Create a git tag with e.g. `git tag -m "Dummy release tag" v5.3.9-wip`
Build with the superbuild as normal, but configure with `-Dome-xml-dir=/path/to/ome-model`.  `ome-model` is directly substitutable for `bioformats`.


pom.xml
```diff
@@ -666,115 +415,80 @@
   </mailingLists>
 
   <scm>
-    <connection>scm:git:git://github.com/openmicroscopy/bioformats</connection>
-    <developerConnection>scm:git:git@github.com:openmicroscopy/bioformats</developerConnection>
+    <connection>scm:git:https://github.com/ome/ome-common-java</connection>
```
sbesson: `ome-model` same below

pom.xml
```diff
@@ -267,76 +90,86 @@
...
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <version>1.9</version>
+          <configuration>
+            <licenseName>bsd_2</licenseName>
+            <organizationName>Open Microscopy Environment:
+    - Board of Regents of the University of Wisconsin-Madison
+    - Glencoe Software, Inc.
+    - University of Dundee</organizationName>
+            <projectName>Common package for I/O and related utilities</projectName>
```
sbesson: Needs to be updated.

dgault:
Testing these locally and I am seeing errors for both Maven and Ant builds. For the ant build it appears that it cannot locate the ome-xml 5.3.0-SNAPSHOt in any of the repos.

Also for the Maven build I am seeing:
[INFO] --- exec-maven-plugin:1.2.1:exec (xsd-fu-metadata-store) @ ome-xml ---
/Users/dgault/Documents/GitHub/rleigh-bioformats/components/ome-xml/target/generated-sources/xsd-fu/ome/xml/meta/MetadataStore.java
/Users/dgault/Documents/GitHub/rleigh-bioformats/components/ome-xml/target/generated-sources/xsd-fu/ome/xml/meta/MetadataRetrieve.java
/Users/dgault/Documents/GitHub/rleigh-bioformats/components/ome-xml/target/generated-sources/xsd-fu/ome/xml/meta/AggregateMetadata.java
/Users/dgault/Documents/GitHub/rleigh-bioformats/components/ome-xml/target/generated-sources/xsd-fu/ome/xml/meta/FilterMetadata.java
/Users/dgault/Documents/GitHub/rleigh-bioformats/components/ome-xml/target/generated-sources/xsd-fu/ome/xml/meta/DummyMetadata.java
Traceback (most recent call last):
  File "../xsd-fu/xsd-fu", line 876, in <module>
  File "../xsd-fu/xsd-fu", line 619, in omeXmlMetadataMain
  File "../xsd-fu/xsd-fu", line 566, in processMetadataTemplate
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/codecs.py", line 896, in open
    file = __builtin__.open(filename, mode, buffering)
IOError: [Errno 2] No such file or directory: '/Users/dgault/Documents/GitHub/rleigh-bioformats/components/xsd-fu/templates-java/OMEXMLMetadataImpl.template'

rleigh-codelibre:
@dgault You're definitely building using this branch: https://github.com/rleigh-codelibre/bioformats/tree/ome-model-split ?  It completely removed xsd-fu and ome-xml from the repo, including the ant logic in ant/toplevel.xml.  There should be zero xsd-fu references in the build.

rleigh-codelibre:
@sbesson Thanks for the comments, going through them now.  I've added excludes for the internal transform files; we can delete/move them later.

Regarding `MetadataConverter` xsd-fu needs to check it to ensure coverage of the API, and that means it can't be in formats-api.  The other classes were already moved, with the formats-api versions extending them.  We should probably formally deprecate them and move all users over to the ome-xml versions.  And repeat this for MetadataConverter.

rleigh-codelibre:
@sbesson The comments should now all be addressed.

sbesson:
Also tested the Sonatype deployment process with `mvn clean deploy`. The `ome-xml`, `specification` and `ome-model` artifacts were successfully uploaded - see https://oss.sonatype.org/content/repositories/snapshots/org/openmicroscopy/. However `xsd-fu` failed with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project xsd-fu: Deployment failed: repository element was not specified in the POM inside distributionManagement element or in -DaltDeploymentRepository=id::layout::url parameter -> [Help 1]
````

We might want to configure the `xsd-fu/pom.xml` to skip deployment as described in http://maven.apache.org/plugins/maven-deploy-plugin/faq.html#skip. This can happen in a follow-up PR.

rleigh-codelibre:
@sbesson I'm thinking it would make sense to remove the xsd-fu pom entirely; all it does is remove the `.pyc` python files.  We could even consider moving the whole thing under ome-xml, e.g. to `src/main/xsd-fu` or `src/xsd-fu`, whatever makes most sense for maven (leaving `src/main/python` for any future generated python bindings).

sbesson:
As `xsd-fu` is minimally a tool for the `generate-sources` phase, migrating it under the top-level `ome-xml` directory certainly sounds sensible. From the [standard directory layout](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) and given `src/main/antlr`, `src/main/xsd-fu` would work.

Is that something we want to investigate in a follow-up PR?

rleigh-codelibre:
I think it would be fine as a followup.  There's already quite a bit in this one!

dgault:
The tests for https://github.com/rleigh-codelibre/bioformats/tree/ome-model-split I could only get to build if I had first built the ome-xml:jar:5.3.0-SNAPSHOT locally but I would assume that is to be expected if the sonatype repo is not being used in https://github.com/rleigh-codelibre/bioformats/tree/ome-model-split

Otherwise this is looking good to merge for me and the builds appear to be working as expected.

sbesson:
:+1: Thanks all. Merging this and pushing the latest SNAPSHOT to the Sonatype snapshots repository. @rleigh-codelibre can you open your Bio-Formats PR consuming https://oss.sonatype.org/content/repositories/snapshots/org/openmicroscopy/ for testing.

PR #2, opened 2016-10-24
Remove xsd-fu POM

sbesson:
xsd-fu is solely used for generating sources in the ome-xml component.
Additionally `mvn clean deploy` fails on this component. In the mid-term, we
might migrate it under `ome-xml` e.g. `src/main/xsd-fu`. For now, this commit
should remove it from being declared in the multi-module components in the
top-level POM.


rleigh-codelibre:
This looks fine, is green and good to merge.

sbesson:
Thanks. Merging and testing a staging release on Sonatype to see if the validation works.

PR #5, opened 2016-10-26
Add badges for Maven Central and Javadocs

sbesson:
See https://github.com/sbesson/ome-model/tree/badges for the review of this PR

Note that I have only linked to the `ome-xml` component but we could add badges for the other components.


hflynn:
Looks nice!

rleigh-codelibre:
Looks good, and good to merge.

I think the ome-xml component is fine for now, particularly if we want to consolidate the components for 6.0.

PR #6, opened 2016-10-31
Fix schema publication script post-migration of the repository

sbesson:
Following the filtering of the specification component, the creation of the intermediate indexes was in a broken state. This commit should restore the index creation.

### DIFF
--- a/specification/publish
+++ b/specification/publish
@@ -306,8 +306,8 @@ createlink() {
     if [ "$4" = "false" ]; then
         reldesc=""
     fi
-    link=$(find published -name "$name" | grep "$release" | head -n1)
-    link=$(echo "$link" | sed -e 's;published/;;')
+    link=$(find $root -name "$name" | grep "$release" | head -n1)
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:
I'm not sure this is the best way to fix things.  Ideally, it would be placed under `target` so it gets properly cleaned.  If we can fix the script to leave it under target, that would be preferable.

sbesson:
@rleigh-codelibre: I can push an extra commit to move to `target`. I might have to fix the SCHEMA-release job and update the schema contributing instructions accordingly.

rleigh-codelibre:
That would be great if it's not too much trouble.

sbesson:
@rleigh-codelibre last commit should revert the `target/` dropping and fix all `find published` instances so that the relative links of the HTML pages should now be generated and work as expected.

rleigh-codelibre:
This all looks fine to me and good to merge.

sbesson:
Thanks. Merging and I'll work through a re-release of the existing schema to test the deployment job

PR #9, opened 2016-11-03
cmake: Drop gtest source building support

rleigh-codelibre:
See https://github.com/ome/ome-cmake-superbuild/pull/103


mtbc:
Good to merge with https://github.com/ome/ome-cmake-superbuild/pull/103.

PR #10, opened 2016-11-11
Enable C++11 support

rleigh-codelibre:
- Require C++11, prefer C++14 if available

Testing: See https://github.com/ome/ome-common-cpp/pull/29


mtbc:
Good to merge when https://github.com/ome/ome-common-cpp/pull/29 goes in.

PR #12, opened 2016-11-24
xsd-fu: Make conversion checks work for C++

rleigh-codelibre:
- Correct the Java checks
- Make the checks work for C++
- Make errors fatal so we catch them (the C++ issues weren't noticed until I started working on the model a few days back)

This adds additional regexes to pick the names from the C++ sources.  It also corrects the logic for determining the conversion filename since this isn't generated, and for C++ we need to distinguish between source and header generation (we run the check only when generating headers since the checks are against the generated headers, plus the static conversion file).  This wasn't an issue for Java.

We also amend the checks to look at the differences between these sets:

```
    (convert) - (store) | (retrieve)
    (store) | (retrieve) - (convert)
```

The existing logic was comparing store and retrieve sets, but we really want to pick up on methods which are in store *or* retrieve, but which are missing in convert.

This change means we have full checking for both the Java and C++ builds at all times.  Anyone changing any of these files will be required to update the interfaces and conversion logic in lockstep for the build to remain green, so this will prevent any inconsistency creeping in.

Testing:

- check jobs remain green
- remove a conversion block from `ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java` and run the maven build (or `mvn generate-sources`).  It should fail with a diagnostic indicating the missing methods during the generation step.
- remove a conversion block from `ome-xml/src/main/cpp/ome/xml/meta/Convert.cpp` and run the C++ build (or just the `gensrc` target).  It should fail with a diagnostic indicating the missing methods during the source generation step.  Basically, run the superbuild with this branch checked out, then run the `ome-model` target.


dgault:
From testing this it is working as described for both CPP and Java. I am still going through the regular expressions as I remember previously that the regex r'(?<=[\s\.:])([gs]et[a-zA-Z0-9]+)' had also returned some unwanted results other than the get and sets, but I haven't found them as of yet.

Just a couple of questions to clarify though:
Is there no need to check for differences between retrieve and store?
Are we sure we want to fail the generation rather than provide a warning?

rleigh-codelibre:
While we *could* check for differences between retrieve and store, in practice one contains only getters and one setters, with a couple of exceptions.  The pairing of the methods is done in the converter and so the comparison with the converter will in practice identify any missing functionality or mismatched getters and setters.

Note see https://github.com/ome/ome-model/blob/master/ome-xml/src/test/cpp/scripts/metadata-convert-coverage.cmake.in which is the original implementation of the checker (in shell).  It's this behaviour which this PR is copying.

As for failing rather than warning, this script has been broken with C++ for months (since it was written) and no one noticed the warnings.  So I think failure is warranted.  Missing functionality in the converter is a serious model bug which will result in dataloss on conversion, and I think treating it as a failure will force any work on the code generation to always ensure correctness in order to compile.  It's certainly a bit harsh, but I don't think warnings are enough, and complying with the check is not that onerous.

dgault:
Completed going through the regex for converter and store and could only find results for get and set so that the expression looks correct ok.

Also tested removing functions from converter and convert.cpp and verifying that error messages are correct with the get message referring to retrieve and the set with store etc. This works as expected.

The logic for displaying errors if a match is in converterDiff:
`for match in convertDiff:
   if match in retrieveSet:
     print("MetadataValidation: %s is present in %s but missing from MetadataStore" % (match, opts.lang.getConverterName()), file=sys.stderr)`

If for example getPixelsSizeX is in convertDiff, it likely wont be in either retrieve or store, as such should the logic check if it contains get or set to determined which file it is missing from?

rleigh-codelibre:
All the names should always contain get or set?  The selection is `([gs]et[a-zA-Z0-9]+)` (previously it was stripped).

dgault:
Sorry I meant should the logic be (pseudo code)

for match in convertDiff:
if match starts with 'set':
print("MetadataValidation: %s is present in %s but missing from MetadataStore" % (match, opts.lang.getConverterName()), file=sys.stderr)
else 
print("MetadataValidation: %s is present in %s but missing from MetadataRetrieve" % (match, opts.lang.getConverterName()), file=sys.stderr)

sbesson:
A couple of general thoughts:
- for future cross-linking reference, the original validation effort was introduced in https://github.com/openmicroscopy/bioformats/pull/2477
- the C++ warnings and the duplication with the cmake implementation were noted during the initial review in https://github.com/openmicroscopy/bioformats/pull/2477#issuecomment-231342632 but pushed from the original PR. Thanks for taking the time to fix them properly
- following the previous comment, will the cmake script now redundant and to be removed in a follow-up PR?
- assuming all methods are covered, I am also in favor of hard failures rather than warnings. We can review this policy as part of the next the model work. For stable branches, being stringent might be a good compromise.

rleigh-codelibre:
Ah, I missed that original PR comment.

I've updated the diagnostic following @dgault's suggestion.  I've also removed the redundant shell script check for C++.

dgault:
The latest commits here look good. Im happy that the new converter checks in the PR are working correctly and agree that having them fail is the correct behaviour.

rleigh-codelibre:
I re-ran the merge trigger after pushing these commits; all the builds are green.

PR #13, opened 2016-11-29
samples: Add Transform example to ROI sample

rleigh-codelibre:
Insight doesn't appear to apply the transform, which is a simple translation.  It validates, but is it correct?


sbesson:
@rleigh-codelibre: thanks for the sample addition. Very useful. As mentioned in https://trello.com/c/InWdSfGf/62-ome-xml-transform-in-bf, the very initial limitation seems to happen at the code-generated Shape class level where the transform element is lost.

/cc @dominikl @mtbc

dominikl:
Thanks @rleigh-codelibre . Yes, the transformation already gets lost in import. But however, even if it wouldn't, Insight doesn't handle transformations properly at the moment.

PR #14, opened 2016-12-01
Make ROI transforms functional

rleigh-codelibre:
- OMEXMLModelObject template:
  - Remove the assumption that the element name will always match the class name;
    this is incorrect for AffineTransform and Transform.
    - Make the OMEModelObject::asXMLElement(Document, Element) method public
      for both C++ and Java.
    - The caller of asXMLElement sets the element name and passes in the element
    - Don't use return value of asXMLElement (Java) and don't return a value at all (C++)
    - Remove the sanity check that the element name and class name match, since
      it is incorrect; for C++ deprecate the validElementName method and always
      return true, for Java remove the inline check.
  - Handle setting the class name differently for abstract (use derived class
    name) and non-abstract (use property name).  This does mean derived classes
    can't use the property name, since they require the name to specify the
    concrete type to instantiate; that's an assumption made by the implementation
    which we now enforce consistently.
  - Remove broken conditional checking for isAbstract.
  - Initialise AffineTransform instances with the identity matrix; this prevents
    serialisation errors when only some of the "Axy" attributes are set, and
    provides a reasonable default when creating one directly.
- Enable the use of createTransform in XMLMockObjects; also set the
  transform to do a basic translation.
- 2008-09-to-2009-09.xsl transform: Don't create an empty "transform"
  attribute, since it's both invalid and breaks the Bio-Formats unit
  tests.

Note: you might find the [whitespace diff](https://github.com/ome/ome-model/pull/14/files?w=1) easier to review for the templates to ignore indentation changes.

Testing:

- Check C++ and Java builds are green [note: does not use regular jobs]
- Using this [transform sample](https://github.com/rleigh-codelibre/ome-model/blob/9e33f33df8fe3223650868e96f0814fa9300c109/specification/samples/2016-06/ROI.ome.xml), run
  - Java: Run `./tools/showinf ROI.ome.xml -nopix -omexml` and you should see a `<Transform>` element in the output (without the PR, this will be missing)
  - C++: Embed the `ROI.ome.xml` into a TIFF with tiffcomment, then run `ome-files info ROI.ome.tiff --omexml` and again you should see a `<Transform>` element which was previously missing

Other testing:

- Is the transform update correct?  Does it still work with transforms from this era (I can't see any obvious samples.)
- Does altering the serialisation order break anything.  The `asXMLElement` order now matches the `update` order, so should be fine, but I'm unsure why this was done differently for different element types in the first place.

### DIFF
--- a/ome-xml/src/main/cpp/ome/xml/model/OMEModelObject.h
+++ b/ome-xml/src/main/cpp/ome/xml/model/OMEModelObject.h
@@ -123,6 +123,8 @@ namespace ome
          * @param name the element name to check.
          *
          * @returns @c true if valid, @c false if invalid.
+         * @deprecated Always returns true since any element name is
+         * now regarded as being valid.
          */
         virtual bool
         validElementName(const std::string& name) const = 0;
@@ -145,6 +147,20 @@ namespace ome
          */
         virtual common::xml::dom::Element
         asXMLElement (common::xml::dom::Document& document) const = 0;
+
+          /**
+           * Transform the object hierarchy rooted at this element to
+           * XML.  This internal implementation of asXMLelement also
+           * requires an XML element, which must not be null, or may
+           * be instantiated and passed from superclasses.
+           *
+           * @param document XML document for element creation.
+           * @param element XML element for setting model data.
+           * @returns an XML DOM tree root element for this model object.
+           */
+        virtual common::xml::dom::Element
+        asXMLElement (common::xml::dom::Document& document,
+                      common::xml::dom::Element&  element) const = 0;
 
         /**
          * Update the object hierarchy recursively from an XML DOM tree.
--- a/ome-xml/src/main/cpp/ome/xml/model/OMEModelObject.h
+++ b/ome-xml/src/main/cpp/ome/xml/model/OMEModelObject.h
@@ -156,9 +156,8 @@ namespace ome
            *
            * @param document XML document for element creation.
            * @param element XML element for setting model data.
-           * @returns an XML DOM tree root element for this model object.
            */
-        virtual common::xml::dom::Element
+        virtual void
         asXMLElement (common::xml::dom::Document& document,
                       common::xml::dom::Element&  element) const = 0;
 
--- a/ome-xml/src/main/cpp/ome/xml/model/OriginalMetadataAnnotation.cpp
+++ b/ome-xml/src/main/cpp/ome/xml/model/OriginalMetadataAnnotation.cpp
@@ -178,12 +178,12 @@ namespace ome
         return false;
       }
 
-      common::xml::dom::Element
+      void
       OriginalMetadataAnnotation::asXMLElementInternal (common::xml::dom::Document& document,
                                                         common::xml::dom::Element&  element) const
       {
 
-        return XMLAnnotation::asXMLElementInternal(document, element);
+        XMLAnnotation::asXMLElementInternal(document, element);
       }
 
       OriginalMetadataAnnotation::metadata_type&
--- a/ome-xml/src/main/cpp/ome/xml/model/OriginalMetadataAnnotation.cpp
+++ b/ome-xml/src/main/cpp/ome/xml/model/OriginalMetadataAnnotation.cpp
@@ -179,13 +179,6 @@ namespace ome
       }
 
       common::xml::dom::Element
-      OriginalMetadataAnnotation::asXMLElement (common::xml::dom::Document& document) const
-      {
-        common::xml::dom::Element nullelem;
-        return asXMLElementInternal(document, nullelem);
-      }
-
-      common::xml::dom::Element
       OriginalMetadataAnnotation::asXMLElementInternal (common::xml::dom::Document& document,
                                                         common::xml::dom::Element&  element) const
       {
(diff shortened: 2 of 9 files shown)


sbesson:
With the latest state of this branch, there are 24 failures in the spec tests of `formats-bsd` (testing upgrade/downgrade with `XMLMockObjects`. I wonder if the issue is that `createTransform` is filling only 2 fields while the schema requires all of them.

rleigh-codelibre:
@sbesson All 6 fields are filled.  When instantiated directly, it defaults to initialising with the identity, and then we are modifying two of those values.

rleigh-codelibre:
@sbesson Tested with maven builds of both, and it's working for me.  I changed:

```
-    <specification.version>5.3.1</specification.version>
-    <ome-xml.version>5.3.1</ome-xml.version>
+    <specification.version>5.3.2-SNAPSHOT</specification.version>
+    <ome-xml.version>5.3.2-SNAPSHOT</ome-xml.version>
```

Did both versions get updated in your builds?  Since it changes the mock objects used for the bioformats test suite in specification, as well as the ome-xml code generation changes, both require updating.

sbesson:
@rleigh-codelibre: ah very good catch. I used https://github.com/sbesson/bioformats/tree/transforms in the merge command which only bumps `ome-xml`. Will push another commit.

dominikl:
Started up a server locally (just easier to check what is imported into the DB then) with the build artifacts mentioned in standup. Imported the ROI.ome.tiff example which contains a transformation and exported the image again. The transformation is preserved through the whole process. 👍

sbesson:
Functionally tested as follows:

- via the Bio-Formats command line tools using the provided OME-XML sample but also a FV1000 fileset which populates `Rectangle.Transform` elements. All works as expected.

- reproduced the workflow from @dominikl by importing the sample OME-XML with various ROI elements including transforms into an OMERO server and exporting the files as OME-TIFF using the various transforms (see [roi.zip](https://github.com/ome/ome-model/files/631571/roi.zip)). For all transforms up to 2010-06, the transform is preserved and for earlier schema versions, the transforms along with several shapes are lost (transformation known to be lossy).

- opened the various generated OME-TIFF using OME Files 0.2.3-12-g784d483 and confirmed all Transform elements are also generated.

- finally re-imported all OME-TIFF back into the server:

````
(ice36) sbesson@lifesci-55368:OMERO.py-5.3.0-m6-75-58e53a7-ice36-b8 $ bin/omero hql 'select count(i) from Image i' -q
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
 # | Col1 
---+------
 0 | 7    
(1 row)
(ice36) sbesson@lifesci-55368:OMERO.py-5.3.0-m6-75-58e53a7-ice36-b8 $ bin/omero hql 'select count(i) from Shape i' -q
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
 # | Col1 
---+------
 0 | 49   
(1 row)
(ice36) sbesson@lifesci-55368:OMERO.py-5.3.0-m6-75-58e53a7-ice36-b8 $ bin/omero hql 'select count(i) from AffineTransform i' -q
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
 # | Col1 
---+------
 0 | 14   
(1 row)
````

Also :+1: for systematically adding `Transform` objects to the `XMLMockObjects`.

Briefly went over the [diff of the Java generated sources](https://ci.openmicroscopy.org/view/Failing/job/OME-MODEL-merge-generated-sources/). Did not spot any obvious issue so far but I could use another pair of eye to go over it as well as the C++ diff /cc @mtbc @dgault @joshmoore 

As discussed with @rleigh-codelibre earlier, given the size and the nature of the changes in generated source like modifying the method access from `protected`  to `public`, once the code-generated diff are cleared, I would vote for releasing these changes alongside the augemented ROI sample as 5.4.0 i.e. a minor release bump.

mtbc:
So it's okay to dump the return value of `super.asXMLElement` and just rely on the side effect?

rleigh-codelibre:
@mtbc Yes, it's returning the `Element` we pass in in this instance, since we're chaining up.

dgault:
Certainly for the Java world I would feel a lot more comfortable keeping the return value of the `super.asXMLElement`, it seems both safer and easier to read and understand. 

Otherwise the diffs look relatively clean ie without any unwanted side effects.

I may have missed the discussion earlier and it may be out of scope for this, but would setting the default values for the AffineTransform not make more sense in the xsd rather than hard coded in the generator?

sbesson:
@rleigh-codelibre is it possible to push a final commit for the return value?

rleigh-codelibre:
I'm unconvinced this needs changing.  The point of asXMLElement returning a value is for inserting child elements when invoking the method on a child model object, and have it create and return the element.  This case isn't a child element; it's chaining up to the superclass, using *its own element*, and as such it has zero value.  If assigning the returned element had any effect, it would be to *overwrite* the content we had already set, which would be a major bug.  Adding this will simply be reassigning the same element *to itself*, which is pointless.  The lack of assignment here is deliberate, to indicate that we don't want to use the return value.

rleigh-codelibre:
Note: There are two overloads of asXMLElement, one with an element parameter and one without.  We no longer require the version with an element parameter to return the same element; the last change makes the return type explicitly ignored throughout the templates, since for every call we create the Element by hand and pass it in explicitly.

rleigh-codelibre:
Since the `asXMLElement` implementation for C++ was private, I've adjusted it to never return anything, including in the new public part of the API.  This means it's completely safe and unambiguous.

dgault:
The important thing here is that the transforms are working correctly and testing indicates that the upgrades are indeed behaving as desired. Altering the serialisation order of the asXML commands also does not appear to have any negative impacts.

PR #15, opened 2016-12-05
Review developers section of the parent POM

sbesson:
Post components decoupling, this PR reviews the developers to use the following conventions discussed with @jburel and @joshmoore:

- lists the team as the main developers (matching the behavior of webapps) and uses the ome-devel email
- lists all contributors of the repository as contributors

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -455,7 +455,7 @@
     <contributor><name>Helen Flynn</name></contributor>
     <contributor><name>David Gault</name></contributor>
     <contributor><name>Alexander Görtz</name></contributor>
-    <contributor><name>Mark Hinerm</name></contributor>
+    <contributor><name>Mark Hiner</name></contributor>
     <contributor><name>Simone Leo</name></contributor>
     <contributor><name>Roger Leigh</name></contributor>
     <contributor><name>Melissa Linkert</name></contributor>
--- a/pom.xml
+++ b/pom.xml
@@ -442,52 +442,28 @@
 
   <developers>
     <developer>
-      <id>hinerm</id>
-      <name>Mark Hiner</name>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -442,52 +442,28 @@
...
+  <contributors>
+    <contributor><name>Chris Allan</name></contributor>
+    <contributor><name>Sébastien Besson</name></contributor>
+    <contributor><name>Jean-Marie Burel</name></contributor>
+    <contributor><name>Colin Blackburn</name></contributor>
+    <contributor><name>Mark Carroll</name></contributor>
+    <contributor><name>Helen Flynn</name></contributor>
+    <contributor><name>David Gault</name></contributor>
+    <contributor><name>Alexander Görtz</name></contributor>
+    <contributor><name>Mark Hinerm</name></contributor>
```
melissalinkert: This should be ```Mark Hiner```

pom.xml
```diff
@@ -442,52 +442,28 @@
...
-      <roles>
-        <role>architect</role>
-        <role>developer</role>
-      </roles>
-      <timezone>-6</timezone>
-      <properties>
-        <picUrl>http://loci.wisc.edu/files/loci/images/people/curtis-2010.jpg</picUrl>
-      </properties>
+      <name>The OME Team</name>
+      <email>ome-devel@lists.openmicroscopy.org.uk</email>
```
sbesson: Some of this is already defined in the `organization` section above in the POM.

mtbc: Aha, true! I wonder if we can simply put an `<organization>` up there to at least name the Open Microscopy Environment but either way this PR is fine to merge as-is.

mtbc:
Aha, now found `<organization>` tag way below. :+1:

PR #16, opened 2016-12-08
Re-order the POM following the code convention

sbesson:
Following the comment from @mtbc in https://github.com/ome/ome-model/pull/15#discussion_r91043917, this PR reorganizes the ome-model POM to follow the Maven ordering convention.

See https://maven.apache.org/developers/conventions/code.html for more details


rleigh-codelibre:
Looks fine to merge to me.

PR #18, opened 2016-12-09
Additional POM cleanup

sbesson:
Following https://github.com/ome/ome-model/pull/16,

- Remove `pluginRepositories` section already defined in the [Super POM](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Super_POM)
- Clean up the module POMs


dgault:
Changes look sensible and the maven build is unaffected. Build runs and completes successfully, all tests pass and artefacts look good. 

PR seems good to merge.

PR #19, opened 2016-12-12
Add sphinx docs

rleigh-codelibre:
Add formats docs from ome-documentation.

- Add docs to `docs/sphinx`
- Rename from `.txt` to `.rst` extensions (to be compatible with cmake)
- Add cmake build
- Add maven build

The goal here is to add the docs and the build logic.  There's some additional cleanup of the sphinx configuration and docs which can happen as a followup.

The cmake build is tested by the superbuild; check the staging docs, and there will be a new ome-model (see also https://github.com/ome/ome-model/pull/20 and https://github.com/ome/ome-cmake-superbuild/pull/109)

- Check docs are built by travis (no linkchecking due to the fragility)
- [jenkins](https://ci.openmicroscopy.org/view/Docs/job/MODEL-DEV-merge-docs/11/); includes downloadable archives
- [staged docs](http://www.openmicroscopy.org/site/support/ome-model-staging/); includes sphinx theme

Testing:

```
cd docs/sphinx
make
make clean
make html
make linkcheck
```

will run maven with the appropriate options.  You'll get a zip file and a tar.gz file containing the build contents.  These are archived as artefacts in the MODEL-DEV-merge-docs job.

Check

- http://www.openmicroscopy.org/site/support/ome-model-staging/ (should contain all formats docs content and theming)
- http://www.openmicroscopy.org/site/support/ome-files-cpp-staging/ (should contain ome-model link with the same content and no theming)


rleigh-codelibre:
Versioning needs fixing up in https://ci.openmicroscopy.org/job/OME-FILES-CPP-DEV-merge-docs/BUILD_TYPE=Release,node=marlin/168/console (use model version not release version, and also set release version correctly)

sbesson:
Discussed the last point with @rleigh-codelibre, this PR also contains a change in the versioning logic of the documentation set (to use the `major.minor.patch` number in addition to the schema). We might need to isolate and review this change. /cc @hflynn 

With regarding to build workflow, the current CI/developer build logic in ome-documentation is defined as follows:

````
cd src/formats
make clean html
make linkcheck
````

which execute the following commands under the hood

````
ant clean -Dsphinx.opts="-W -D html_theme=sphinx_theme"   -Dsource.branch="develop/merge/daily" -Dsource.user="snoopycrimecop" 
ant html -Dsphinx.opts="-W -D html_theme=sphinx_theme"   -Dsource.branch="develop/merge/daily" -Dsource.user="snoopycrimecop" 
ant linkcheck -Dsphinx.opts="-W -D html_theme=sphinx_theme"   -Dsource.branch="develop/merge/daily" -Dsource.user="snoopycrimecop" 
```

What are the exact new commands/arguments one should use to reproduce these steps with the current proposal? Should we start enabling a staging docs jobs for building and deployment the staging documentation for review?

rleigh-codelibre:
@sbesson All the missing functionality should now be added.  There's a minor issue in the theme usage, but this is likely a tweak needed in the jenkins job.

hflynn:
I think the staging docs are still being hosted from the old job - I'm not seeing the xsd-fu link fix on http://www.openmicroscopy.org/site/support/ome-model-staging/developers/index.html

rleigh-codelibre:
@hflynn The old job has been disabled, but I don't know if there's some additional magic required here.

hflynn:
I would suspect so, because I've rechecked on a different browser and it's definitely not a caching issue so your new job obviously isn't over-writing the content on web-prod or wherever it is hosted.

hflynn:
N.B. I'll open a PR against contributing as well as one to delete these docs from ome-documentation.

rleigh-codelibre:
@hflynn The versioning is intentional, but certainly something we can change back if needed.

The reason for this is that the documentation is now built by cmake and maven as part of the C++ and Java builds, and packaged up and distributed as part of the normal release process.  As we move the C++ and Java samples into the documentation, this documentation will be documenting the specific version of the ome-model codebase it exists within, and so is versioned accordingly.  That is to say, it's no longer the "2016-06" model version, it's the 5.3.2 ome-model (specification and ome-xml) version which defines and implements the 2016-06 model.  It brings the documentation into the normal release process since it's not just existing in this repository, it's an integrated part of the codebase.

This is not to say we can't build and use the docs independently, but that it's generally a bad idea.  This means we can update the code samples and the rest of the documentation at the same time as the code changes which it's documenting.  That has been a major historical problem with the documentation process as it stands, and migrating it into this repository alongside the code allows us to rectify that.

rleigh-codelibre:
Updated to add model version to title; now staged.

hflynn:
Staging is looking good now. I don't know whether you think the schema version AND the model version in the docs title is overkill/confusing or helpful @sbesson - I'm not sure. Certainly we should add to the first sentence to explain the new versioning but that doesn't need to be on the PR and can happen when the index page is redone.

sbesson:
I agree that `OME 2016-06 Data Model and File Formats 5.3.2-SNAPSHOT Documentation` is confusing. I would streamline that down to a single version number. I think expanding the first paragraph to explain the relationship between the specification version and the model version would suffice and I agree this does not need to happen here. Should we revert 9830d27 then?

hflynn:
Yes, I think revert. Thanks for trying it @rleigh-codelibre but on reflect, I think it's too much noise

rleigh-codelibre:
OK, I've reverted it, but kept the `model_version` substitution in place for future use; we can add this to the first paragraph or something.

hflynn:
Yes, having it to use in the first paragraph is great, saves us having to remember to manually update it with releases 👍

hflynn:
I think this PR is probably good to merge now. We can update the existing content elsewhere now it's all in the right place.

docs/sphinx/conf.py
```diff
@@ -24,7 +24,7 @@
 # -- General configuration -----------------------------------------------------
 
 # General information about the project.
-project = u'OME Data Model and File Formats'
+project = u'OME ' + model_version + u' Data Model and File Formats'
```
sbesson: With the last revert the project name is now mismatching with the index page. Should we drop `model_version` here too?

rleigh-codelibre: Yes, that was an oversight.  Pushed a fix now.

hflynn:
I would favour putting the version history file in place when redoing the index landing page.

rleigh-codelibre:
Sounds fine to me.

docs/sphinx/pom.xml
```diff
@@ -0,0 +1,169 @@
...
+    <groupId>org.openmicroscopy</groupId>
+    <artifactId>ome-model</artifactId>
+    <version>5.3.2-SNAPSHOT</version>
+    <relativePath>../..</relativePath>
+  </parent>
+
+  <groupId>org.openmicroscopy</groupId>
+  <version>5.3.2-SNAPSHOT</version>
+
+  <artifactId>ome-model-docs</artifactId>
```
sbesson: As I was going through the release process of ome-model 5.4.0 and reviewing the artifacts, one question on the naming scheme following https://github.com/openmicroscopy/bioformats/pull/2701#issuecomment-268545866. Before making the final commitment, should we modify the artifact id to use the same grammatical form across the board @rleigh-codelibre ?

rleigh-codelibre: I think the plain `-doc` form would be nicer to use if that's acceptable.

sbesson: Using the `-doc` suffix across the board works for me. @rleigh-codelibre: could you open a PR with the new proposal?

rleigh-codelibre: Shall I update this PR accordingly?  Which other components need tweaking, bioformats?

sbesson: This PR is already merged, we will need a new one before we cut ome-model 5.4.0.

PR #20, opened 2016-12-14
build: Change cmake project name to ome-model

rleigh-codelibre:
This is to match the maven project name and give the sphinx manual
the correct name.

See https://github.com/ome/ome-cmake-superbuild/pull/109 for testing instructions.


simleo:
Good to merge since #109 is OK.

PR #21, opened 2016-12-27
cmake: FindBoost: Add support for Boost 1.63

rleigh-codelibre:
See https://github.com/ome/ome-cmake-superbuild/pull/110


sbesson:
Merging as per https://github.com/ome/ome-cmake-superbuild/pull/110#issuecomment-270343282

PR #22, opened 2017-01-04
build: Use -doc rather than -docs in artefact name

rleigh-codelibre:
Testing: check artefact name, now `ome-model-doc-5.3.2-SNAPSHOT-manual.*`


sbesson:
:+1: Merging for ome-model 5.4.0 release.

PR #23, opened 2017-01-04
Fix intermittent hyperlinks

sbesson:
The model docs jobs have been failing the linkcheck step regularly in the last few days. This PR tries to mitigate the issue by:

- removing the link to the full specification documentation for LSID. The PDF is now restricted to members only and not public
- using HTTPS instead of HTTP in a few places

### DIFF
--- a/docs/sphinx/developers/id-and-lsid.rst
+++ b/docs/sphinx/developers/id-and-lsid.rst
@@ -14,12 +14,7 @@ concept. It was designed to allow the naming or identifying of data and
 associated metadata that can be stored in multiple, distributed data
 stores.
 
-For further information see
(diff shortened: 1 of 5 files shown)


sbesson:
With the latest version of this PR, https://ci.openmicroscopy.org/view/Failing/job/MODEL-merge-docs/26/ is green and I would expect a reduction of the intermittent failure of the https://ci.openmicroscopy.org/job/OME-FILES-CPP-DEV-merge-docs/ jobs as well /cc @rleigh-codelibre

rleigh-codelibre:
The replacement links are all correct, and so all look fine to me.  Let's see if it improves things over the next few days.

PR #26, opened 2017-01-30
Fix permanent redirects

hflynn:
This updates all the links which are being permanently redirected* by the merge build (https://ci.openmicroscopy.org/view/Docs/job/MODEL-merge-docs/) - builds should be green with only our own links redirected to current versions*) and the staging pages should be unchanged.

N.B. I've replaced the TIFF link altogether because the Adobe link was redirecting somewhere irrelevant and is another site which wants you to identify your country of origin before it'll point you at links - hopefully the wiki is sufficient.

*with the exception of the GELifeSciences link which wants you to pick a country so doesn't play nicely with the linkchecker

### DIFF
--- a/docs/sphinx/developers/EnumTool.rst
+++ b/docs/sphinx/developers/EnumTool.rst
@@ -141,5 +141,5 @@ which Enum Tool makes heavy use of internally.
 
 .. SeeAlso::
 
-    - `http://genshi.edgewall.org/ <http://genshi.edgewall.org/>`_
+    - `https://genshi.edgewall.org/ <https://genshi.edgewall.org/>`_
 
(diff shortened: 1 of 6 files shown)


sbesson:
Barring the current unavailability of http://bigtiff.org/ which breaks the builds, these changes all look fine. One addition I would carry on as part of this PR is the Genshi link

```
(line  144) redirect  http://genshi.edgewall.org/ - with Found to https://genshi.edgewall.org/
```

The TIFF Wikipedia page includes several links for the specification and resources so I think it is a valid starting point unless there is a better canonical URL /cc @rleigh-codelibre

hflynn:
@sbesson thanks for spotting that

sbesson:
https://ci.openmicroscopy.org/view/Failing/job/MODEL-merge-docs/49/ and https://ci.openmicroscopy.org/view/Failing/job/OME-FILES-CPP-DEV-merge-docs/225/. :+1:

PR #27, opened 2017-02-06
Revert metadata model changes

rleigh-codelibre:
See https://trello.com/c/DmqJO7bn/64-revert-initial-omemodel-extension-pr

This reverts https://github.com/ome/ome-model/pull/17 so we can release ome-files 0.3.0 without it.  We'll revert the revert later.

Since it preceded the C++11 changes, it's not identical to the revert, it reverts the ome model changes but retains the C++11 changes.  I've taken care to ensure it's not regressed when fixing up the conflicts, but there's a small chance something might have crept in.

Testing: Check all builds are green.


rleigh-codelibre:
@sbesson For the public `ome::compat` to `std` renaming, while it's counterintuitive this is *not* actually a breaking change.  The `ome::compat` types were *already* being defaulted to use `std` internally so in practice there isn't a change here.  The point of the typedefs is to abstract the implementation, and the change here is to expose the standard types which in practice were already being used (as evidenced by the fact that ome-files continued to build with no changes).

The backed out changes certainly would be breaking though; maybe we should do 6.0.0 for them?

rleigh-codelibre:
@sbesson Removed MetadataModel and OMEXMLMetadataModel.  Note they were completely unused and not compiled or installed, so this is simply a final removal from git.

PR #28, opened 2017-02-07
ome-xml: Performance improvements

rleigh-codelibre:
See [trello card](https://trello.com/c/WR9yB2L9/84-metadata-performance) for more details.

`getTagName()` already returns a tag without a namespace, making this pointless.  Removing the unnecessary stripping is a minor performance enhancement.

The major change here is to remove all use of xerces DOMNodeList (ome::common::xml::dom::NodeList).  This has very inefficient iterator performance--O(n) for every indexed access.  Swapping this with the lower level Node `getFirstChild()` and `getNextSibling()` results in a dramatic speedup (around 2 mins to 4 seconds for the mitocheck data!)

Testing: Check builds remain green.

Check the performance of the ome-files-performance tests; they should be significantly faster, for mitocheck in particular.

Test data: download necromancer:~rleigh/ptest.tar.gz
Run `ome-files info` on the two images.  Without this PR merged you'll get very poor performance; with the PR merged you'll get decent performance.

### DIFF
--- a/ome-xml/src/main/cpp/ome/xml/model/OriginalMetadataAnnotation.h
+++ b/ome-xml/src/main/cpp/ome/xml/model/OriginalMetadataAnnotation.h
@@ -46,7 +46,6 @@
 #include <ome/common/xml/dom/Document.h>
 #include <ome/common/xml/dom/Element.h>
 #include <ome/common/xml/dom/Node.h>
-#include <ome/common/xml/dom/NodeList.h>
 
 #include <ome/xml/model/XMLAnnotation.h>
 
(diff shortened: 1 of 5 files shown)


mtbc:
`ome-files info` on `BBBC/NIRHTa-001.ome.tiff` and `mitocheck/00001_01.ome.tiff`` takes a total of 10s with the latest build from `cod` and over 50s with the previous build. Not sure if we're still waiting for `sole` on OME-FILES-CPP-DEV-merge-superbuild to turn green.

mtbc:
Diff looks good. :+1:

sbesson:
Retested using the metadata benchmarking scripts against the BBBC and the mitocheck OME-TIFF. For reference, using OME Files 0.2.3:

metadata-0 2 3

and with this PR included:

metadata read-0 3 0-dev

The additional differences with the Java implementations might require additional investigation but the performance gain with this PR included is significant. Including for the next release.

PR #29, opened 2017-02-09
cmake: Obtain release version from maven

rleigh-codelibre:
Rather than obtain the version from git as done historically, parse the version from the maven POM.

I also deleted some unused cruft which is a legacy from the decoupling.

Add the same source release target as ome-common, so the superbuild was also updated to use this (see https://github.com/ome/ome-cmake-superbuild/pull/121).  Check the source artefacts for ome-model.

Testing: Will continue to build as before, but the version will have changed from 5.3.1 (from the `git describe` output `v5.3.1-nnn-gxxxxxx`) to 6.0.0 from the POM.  This will be displayed in the build log.  The library version will also have changed to 6.0.0 (PRs against ome-files and ome-qtwidgets bump their minimum requirement since it's a breaking version increment).


rleigh-codelibre:
The genshi thing crept in by accident, removed now.
Version updated to 5.5.0-SNAPSHOT

pom.xml
```diff
@@ -7,7 +7,7 @@
 
   <groupId>org.openmicroscopy</groupId>
   <artifactId>ome-model</artifactId>
-  <version>6.0.0-SNAPSHOT</version>
+  <version>5.5.0-SNAPSHOT</version>
```
sbesson: As this is a multi-module component, the versions in all the POMs should be updated

rleigh-codelibre: Done

sbesson:
All builds are green. Merging for upcoming 5.5.0

PR #37, opened 2017-05-24
Bump OMECommon and OMECompat requirements in CMakeLists.txt

sbesson:
Prior to releasing `ome-model 5.5.3` I realized that the upstream dependencies had not been bumped previously. Our usage of the superbuild hides the review of this part as the latest/last release of the upstream libraries are always consumed.

I would assume either `5.4.0` should be the minimal required version of `ome-common` as it includes the C++ 11 work or `5.4.1` as this version is required to build with Boost 1.64 which is supported in `ome-model 5.5.3`. Thoughts @rleigh-codelibre ?


rleigh-codelibre:
I think 5.4.0 is fine as a minimum; we haven't changed the minimum Boost version, so that's not a consideration in this case (we only added support for 1.64, it's not mandated and we don't use any 1.64-only features).

rleigh-codelibre:
Thanks.  Tested locally and it's fine.  This is good to merge.

This will need repeating for the ome-files-cpp and ome-qtwidgets repositories.

sbesson:
https://ci.openmicroscopy.org/view/Failing/job/OME-FILES-CPP-DEV-merge-trigger/553/ and all subjobs being green, merging and I'll work through the steps of the 5.5.3 release later today. @rleigh-codelibre could you take care of the bump of the `ome-files-cpp` and `ome-qtwidgets` dependencies?

PR #38, opened 2017-06-06
Update API reference including use of deleted methods

rleigh-codelibre:
Upgrade to the latest Doxygen configuration file version.

Doxygen can't handle deleted methods too well--it complains about
them not being documented despite them not existing at all, so
skip them entirely.

Testing: Check all builds remain green.

### DIFF
--- a/docs/doxygen/ome-xml.dox.cmake
+++ b/docs/doxygen/ome-xml.dox.cmake
@@ -1,112 +1,129 @@
-# Doxyfile 1.8.4
+# Doxyfile 1.8.13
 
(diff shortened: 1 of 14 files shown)


dgault:
Similar to comments in https://github.com/ome/ome-files-cpp/pull/70, the new and extended additions all look fine, as does the handling of the deleted copy constructor and assignment operator with the `= delete` and SKIP block.

The doxygen settings file appears to have updated correctly and none of the existing settings have been modified.

Builds and tests are green so this looks good to merge.

PR #39, opened 2017-06-06
cmake: Use C++ standard variables as documented

rleigh-codelibre:
See [trello card](https://trello.com/c/0JzeUNqd/38-correct-use-of-cmake-standard-version-selection-variables).

Testing: check builds are green; there's no behaviour change here.


dgault:
Only change is the removal of the CMAKE_CXX_STANDARD_REQUIRED, as mentioned on https://github.com/ome/ome-common-cpp/pull/40 this change should not have any impact on exitsing behaviour.

All builds and tests are green so this PR should be good to merge.

PR #41, opened 2017-06-11
xsd-fu: Correct C++ model object template using wrong element

rleigh-codelibre:
[Trello](https://trello.com/c/S7fIcUi9/7-correct-xsd-fu-template-mistake)

Compare with the Java template, line 433.  We make it use each element for the NodeList we are iterating over, rather than the current element.

Testing: Looking at the generated sources, this codepath appears to be unused at present.  Maybe simply confirming we match the Java behaviour is sufficient to check?


dgault:
This change looks to be correct from my perspective and matches the Java behaviour on https://github.com/ome/ome-model/blob/master/xsd-fu/templates-java/OMEXMLModelObject.template#L433

PR #42, opened 2017-06-27
cmake: Move Options inclusion after project call

rleigh-codelibre:
Options.cmake uses find_package to find Doxygen.  This uses
PROJECT_BINARY_DIR, which isn't set until after project() is called.
FindDoxygen in cmake 3.9.0 breaks due to using PROJECT_BINARY_DIR.

See https://github.com/ome/ome-cmake-superbuild/pull/146 for testing instructions.


sbesson:
See https://github.com/ome/ome-cmake-superbuild/pull/146#issuecomment-313358701

PR #44, opened 2017-07-04
Use new OME sphinx theme

rleigh-codelibre:
Same as https://github.com/openmicroscopy/ome-internal/pull/353 but for ome-model.

It also fixes up the conf.py syntax to build with Python 3 sphinx (still works with Python 2).


sbesson:
screen shot 2017-07-07 at 14 53 36

Merging

PR #49, opened 2017-08-23
cmake: FindBoost requires cmake 3.4

rleigh-codelibre:
See https://github.com/ome/ome-cmake-superbuild/pull/159 for testing instructions.


sbesson:
See https://github.com/ome/ome-cmake-superbuild/pull/159#issuecomment-325378479

PR #50, opened 2017-08-24
Add versioning note & update README

hflynn:
Discussed with Seb earlier that the new versioning of the Model docs is a bit confusing because the only versioning discussed within the docs is the Schema versioning which is different, so I've added a note to try and clarify if anyone else is confused.

While I was here I noticed I'd missed the docs links in the README when I updated them in the docs. Note that I can't update the C++ API docs as they haven't been migrated, as noted on https://trello.com/c/aKaDecyV/78-link-to-c-api-docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ OME data model specification, code generator and implementation.
 Links
 -----
 
-- [Documentation](http://www.openmicroscopy.org/site/support/ome-model/)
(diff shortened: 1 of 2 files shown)


mtbc:
Reviewing based on https://github.com/hflynn/ome-model/tree/versioning-note and https://ci.openmicroscopy.org/job/MODEL-merge-docs/222/.

PR #51, opened 2017-08-30
Boost 1.53 revert

rleigh-codelibre:
See https://trello.com/c/YJBB1cOA/11-centos-7-platform-support

Not for merging at this point.


sbesson:
See https://github.com/ome/ome-common-cpp/pull/49#issuecomment-328029320

PR #52, opened 2017-09-08
cmake: FindBoost: Add support for Boost 1.65.1

rleigh-codelibre:
See https://github.com/ome/ome-common-cpp/pull/50


sbesson:
See https://github.com/ome/ome-common-cpp/pull/50#issuecomment-329510000

PR #56, opened 2017-10-10
Fix & update links

hflynn:
The old http://www.svisionllc.com/ link isn't resolving to anything but https://www.drvtechnologies.com looks to be the right company (google has cached results to link them together as well as the name being right).

Also fixed the remaining plone link that is now migrated to the new website and updated the downloads links for BF & OME Files to just use the new pages on the website as they only need to link to the latest version anyway, and readme links to use https for downloads, website and docs.

### DIFF
--- a/docs/sphinx/ome-xml/java-library.rst
+++ b/docs/sphinx/ome-xml/java-library.rst
@@ -12,7 +12,7 @@ Download
 
 **ome-xml.jar** is included in **bioformats_package.jar** available to
 download from the :bf_downloads:`Bio-Formats download page <>`. Alternatively,
-the individual artifact **ome-xml.jar** is hosted at `Maven Central <http://search.maven.org/`_ under the group ``org.openmicroscopy``. The
+the individual artifact **ome-xml.jar** is hosted at `Maven Central <http://search.maven.org/>`_ under the group ``org.openmicroscopy``. The
 license is `LGPL <http://www.gnu.org/licenses/lgpl.html>`_.
 
 Installation
--- a/docs/sphinx/ome-xml/java-library.rst
+++ b/docs/sphinx/ome-xml/java-library.rst
@@ -10,8 +10,9 @@ metadata processing facilities form the backbone of the
 Download
 --------
 
-You can download **ome-xml.jar** from the 
-:bf_downloads:`Bio-Formats downloads <>` page. The
+**ome-xml.jar** is included in **bioformats_package.jar** available to
+download from the :bf_downloads:`Bio-Formats download page <>`. Alternatively,
+you can download **ome-xml.jar** as a single jar from `Maven <http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.openmicroscopy%22%20AND%20a%3A%22ome-xml%22>`_. The
 license is `LGPL <http://www.gnu.org/licenses/lgpl.html>`_.
 
 Installation
--- a/docs/sphinx/ome-xml/java-library.rst
+++ b/docs/sphinx/ome-xml/java-library.rst
@@ -12,7 +12,7 @@ Download
 
 **ome-xml.jar** is included in **bioformats_package.jar** available to
 download from the :bf_downloads:`Bio-Formats download page <>`. Alternatively,
-you can download **ome-xml.jar** as a single jar from `Maven <http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.openmicroscopy%22%20AND%20a%3A%22ome-xml%22>`_. The
+the individual artifact **ome-xml.jar** is hosted at `Maven Central <http://search.maven.org/`_ under the group ``org.openmicroscopy``. The
 license is `LGPL <http://www.gnu.org/licenses/lgpl.html>`_.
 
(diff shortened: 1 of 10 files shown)


rleigh-codelibre:
All looks fine to merge to me.  While today's build failed, all the links are fine.

hflynn:
The build doesn't like the drvtech URL or the Maven one though, I don't want to merge this if it's going to just start breaking the latest build too :(

The Maven error is an anchor not found but the drvtech one looks to be something to do with the request type:
`<urlopen error [Errno 1] _ssl.c:492: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error>`

Any ideas @sbesson ? Is this maybe something that upgrading Sphinx would help? Could we do that on this repo given the issue is incompatibility with OMERO 5.3.x? Or is this build actually using a later sphinx anyway, I can't find the version given the docs are set up differently on this repo.

rleigh-codelibre:
It working perfectly for me with Sphinx 1.5.6.  Now the OMERO release is out of the way, can we upgrade all our build nodes to use the latest Sphinx?

sbesson:
The main reason for which Sphinx was not upgraded on our build nodes was the need to release OMERO 5.3.x (OMERO 5.3.5 in the most recent case) where Sphinx > 1.2.43 is not supported. As the Sphinx dependency is shared across all nodes, upgrading it means we would have to workaround some deployment issues in the case of a security 5.3.x release for building the documentation. Some workarounds I can think of:

- temporarily downgrade the Sphinx to 1.2.3 for the time of the security release
- and/or isolating the OMERO release documentation jobs in an environment where Sphinx can be capped

In short, we can certainly make this decision and upgrade Sphinx on CI and it would make sense with regard to the efforts made at the code level across the board. We mostly need to collectively agree on the implications mentioned above /cc @jburel @joshmoore

rleigh-codelibre:
Can't we create an alternative virtualenv for use with the old 5.3 jobs, and update the default ones?

hflynn:
For the OMERO docs, there isn't a specific 5.3 job (or branch) @rleigh-codelibre which is part of the problem - the 5.3.5 release was done by temporarily pointing the DEV job at a tag

docs/sphinx/ome-xml/java-library.rst
```diff
@@ -10,8 +10,9 @@ metadata processing facilities form the backbone of the
 Download
 --------
 
-You can download **ome-xml.jar** from the 
-:bf_downloads:`Bio-Formats downloads <>` page. The
+**ome-xml.jar** is included in **bioformats_package.jar** available to
+download from the :bf_downloads:`Bio-Formats download page <>`. Alternatively,
+you can download **ome-xml.jar** as a single jar from `Maven <http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.openmicroscopy%22%20AND%20a%3A%22ome-xml%22>`_. The
```
sbesson: Interestingly, the anchor was not found in https://ci.openmicroscopy.org/view/Docs/job/MODEL-merge-docs/262/console although the link resolves to me.
I wonder if people will download the JAR manually. If a Java project wants to consume the ome-xml Java library, they will probably use a dependency management tool (Maven, Gradle, Ivy). It might be sufficient to say the artifact is hosted Maven Central (with a link to https://search.maven.org/) under the group `org.openmicroscopy` and the name is `ome-xml`.

hflynn:
@sbesson update pushed for the Maven anchor

hflynn:
Found a few more old URLs so pushed a fix for those too

hflynn:
We mention Maven extensively in the BF docs without linking to it and it is easy to find so I'm going to remove the offending link here and then hopefully we'll be able to upgrade to Sphinx 1.4.9 is this morning and get this green.

rleigh-codelibre:
The latest change looks OK, but I would rather we upgrade sphinx and have proper links to maven central.

hflynn:
Re-running the merge build including this PR with Sphinx upgraded to 1.4.9

hflynn:
Warning, treated as error:
/opt/hudson/workspace/MODEL-merge-docs/src/docs/sphinx/ome-tiff/index.rst:37: WARNING: broken link: https://www.drvtechnologies.com

Still not coping with the timelag on this link :(

hflynn:
If we don't want to add this link to the ignore list, maybe we should exclude this PR until we have a plan for updating the build?

hflynn:
This is finally green now that the job has been upgraded to Python 2.7 on the CI. I'll try reinstating the Maven link and see if it stays that way!

hflynn:
Hah! Now the job is red again because Zeiss.com isn't loading - the Maven link is fine though so this should be good to merge finally!!

hflynn:
The Maven link passes locally with Sphinx 1.6.3

hflynn:
Cleaned up unneeded commits removing and adding back the Maven link - this should be good to merge now the build is upgraded and everything is green.

sbesson:
https://ci.openmicroscopy.org/view/Docs/job/MODEL-merge-docs/ has been steadily green for the last few days.

PR #59, opened 2017-11-27
cmake: Remove find modules distributed with CMake

rleigh-codelibre:
See https://github.com/ome/ome-common-cpp/pull/53


joshmoore:
Merging on :+1: from https://github.com/ome/ome-common-cpp/pull/53#issuecomment-347862867

PR #60, opened 2017-11-30
Fix GE Healthcare Life Sciences link

hflynn:
This link was breaking the build this morning - see https://ci.openmicroscopy.org/view/Failing/job/MODEL-merge-docs/312/console and their website has a banner saying they are upgrading it so I've gone for just linking to the main page for now - there's no nice Deltavision landing page at the moment and I don't want to use a URL that then gets broken again as they upgrade.


hflynn:
https://ci.openmicroscopy.org/view/Docs/job/MODEL-merge-docs/313/ is green again with this included

PR #65, opened 2018-01-10
Add support for building with Python 3 as well as Python 2

rleigh-codelibre:
This is primarily intended to allow building and testing of the C++ code and ome-files-py with Python 3, though it will also permit building of the ome-model Java code with Python 3 as well.  This is the first step for moving to Python 3 as discussed a few weeks back, and is a proof of concept for the migration of other Python code.

- A patch to from Genshi upstream SVN has been applied to fix some Python 3 bugs in Genshi.  Since upstream is dead, these fixes will likely never be in a new public release making incorporation into our fork the only viable option.
- generateDS was updated to fix StringIO with Python 3 (patch taken from the generateDS upstream)
- All python code has been upgraded with `python-modernize` ([docs](https://python-modernize.readthedocs.io/en/latest/)).  Done by running `python-modernize -w <file>` for each source file with Python 3.6.3.  This upgrades the code to Python 3, and then adds the necessary compatibility imports to make it work with Python 2.  This means the code will continue to work with Python 2 and Python 3 simultaneously, to allow for a non-breaking transition to Python 3.
- The C++ and Java code templates required manual conversion since the Python code is embedded.  This was primarily using lists in place of iterators and sorting dictionary keys, to retain the Python 2 behaviour of the code.
- The `configobj` module used to load the enum mapping overrides wasn't easily upgradable to Python 3.  I moved the configuration directly into `config.py` alongside the other configuration and stripped out `configobj` entirely.
- `util.odict` also stripped out and replaced by `collections.OrderedDict` which is in both 2.7 and 3.x (`odict` is legacy from earlier 2.x releases).
- Custom template content explicitly converted from bytes to strings.

Now, a small downside.  The key/value ordering of `dict` is implementation-defined, and CPython 2.7 and 3.x differ.  This means it's not possible to preserve the emitted code ordering (e.g. of class members and methods).  I've used `OrderedDict` in our generator and generateDS to ensure a consistent ordering from now onwards (the *order declarared in the XSD schema file*).  But this means a break from the ~random-but-consistent-for-2.x order of 2.7 to declaration order.  This will make the diffs between the old and new generator rather large, but there are no actual code changes except for reordering of things.

This change also drops Python 2.6 support, and will require the model jobs to be moved onto a CI build node with Python 2.7, or the use of SCL Python 2.7.  Tested Python releases:

- 2.7
- 3.6
- 3.7

Hopefully this will be instructive as an example of how Python 2 to 3 code conversion can work on a small scale, which will be applicable to other OME components.

Testing:

- Check C++ and Java builds remain green for Travis and Jenkins.
- Check code generation differences with and without this PR, using the [generated-source-changes](https://ci.openmicroscopy.org/job/OME-MODEL-merge-generated-sources/538/) job, or the finer-grained changes linked to from https://github.com/ome/ome-model/pull/65#issuecomment-421641032.  While there are a lot of rearrangements of the order of functions, there should be no changes to the code itself.

/cc @dgault


sbesson:
An immediate impact of this PR are the set of failures for some of the OME CI jobs on different operating systems. Addressing these require infrastructure changes for a few jobs/nodes as carded independently. Given the ongoing priorities including an ome-model patch release, excluding this PR to fix the builds and get us back to a releasable state.

Let us table this background investigation for now until we have a chance to review the expectations and the scope of the work at an upcoming Wednesday discussion. We can then decide if integration branches would help driving it without affecting other's work. On the CI front, if larger changes are required, this might be a good occasion to start porting some of these jobs to devspace or similar solutions where we have more development & deployment flexibility.

rleigh-codelibre:
@sbesson @joshmoore Now bio-formats 5.8.1 is done, could we get this reviewed and merged?  This will enable us to start testing C++ builds with python3.

rleigh-codelibre:
@sbesson @joshmoore Would it be possible to review this now?  The yaclifw blocker is resolved, and it's fully tested by travis.

rleigh-codelibre:
Current generated source changes are here: https://ci.openmicroscopy.org/job/OME-MODEL-merge-generated-sources/538/

rleigh-codelibre:
I have re-tested with Python 3.6 and the newly-released 3.7.  There are no code generation differences, and the build and tests complete successfully in both cases.

I would find it very useful to start using this functionality.  Would it be possible to review and merge this in the near future?  It's been open for over six months.

rleigh-codelibre:
Copied to [GitLab](https://gitlab.com/codelibre/ome-model/merge_requests/90)

The [GitLab pipeline](https://gitlab.com/rleigh/ome-model/pipelines/30112326) handles automatic generated source diff checking as part of the merge request workflow, and allows direct download of the source changes for review.

rleigh-codelibre:
[GitLab !93](https://gitlab.com/codelibre/ome-model/merge_requests/93) is a test and review of the main reordering changes in this PR, to validate that there are no functional changes and that all reordering is to align the code with the schema order.

rleigh-codelibre:
On GitLab, I have split the changes into several merge requests so that the generated source changes can be reviewed for each commit which changes the generated sources:

- [xsd-fu: generateDS: Preserve element and attribute ordering](https://gitlab.com/codelibre/ome-model/merge_requests/93)
- [xsd-fu: Make templates compatible with Python 3](https://gitlab.com/codelibre/ome-model/merge_requests/94)
- [xsd-fu: Replace odict with OrderedDict from collections](https://gitlab.com/codelibre/ome-model/merge_requests/95)
- [xsd-fu: Modernise genshi and generateDS to work with Python 3 or Python 2.7](https://gitlab.com/codelibre/ome-model/merge_requests/96)
- [xsd-fu: Use stable metadata store method ordering](https://gitlab.com/codelibre/ome-model/merge_requests/97)
- [xsd-fu: Make variable diagnostic strings debug output](https://gitlab.com/codelibre/ome-model/merge_requests/98)
- [Add support for building with Python 3 as well as Python 2](https://gitlab.com/codelibre/ome-model/merge_requests/90)

For each of the merge requests, if you look at the latest pipeline, click on the "Changes" jobs and then "Download" you can get the generated source changes diff (yellow for changes, green if no change).  Or if you go to the "Generator" jobs, you can download the full generated sources and do the diff yourself.  This should make it easier for you to review the rather large set of changes in isolation for each change made, and be confident it's doing what it's supposed to.

This should also serve as an example of a more complex GitLab CI workflow, which is a significant improvement over the Jenkins and Travis workflows it supersedes.

rleigh-codelibre:
I've pushed the rebased set of changes which was tested and merged on GitLab so that you can see an equivalent set of changes.  Or split them up and merge separately if you wish.

sbesson:
Excluding as mentioned in https://github.com/ome/ome-model/pull/84#issuecomment-423169697 to get the version fix out.

rleigh-codelibre:
@sbesson Can be re-enabled whenever convenient.  With #84 merged, I've rebased it onto master (no changes).

rleigh-codelibre:
I have also included Python 3.7 support with the single commit from [!114](https://gitlab.com/codelibre/ome-model/merge_requests/114).

rleigh-codelibre:
Note that in the interim I have also tested and merged
- [!143](https://gitlab.com/codelibre/ome/ome-model/merge_requests/143) - Python3 native
- [!144](https://gitlab.com/codelibre/ome/ome-model/merge_requests/144) - Python3 cleanup
They can also be added here should you want them, or can be opened as separate pull requests once this one is merged.

rleigh-codelibre:
You might also want to consider [!150](https://gitlab.com/codelibre/ome/ome-model/merge_requests/150) which adds support for Python 3.8.

rleigh-codelibre:
@sbesson @joshmoore Would you like this PR updating with the Python 3.8 support linked to above?  Or would you like to merge what's in this PR first, and handle that separately?

joshmoore:
Hi @rleigh-codelibre. Since this has been in the build for some time, let's get it sorted before adding further changes. Thanks for the heads up.

rleigh-codelibre:
@sbesson To answer your questions:

- Yes, the six dependency can be completely dropped, assuming you're happy with ome-model being Python 3-only going forward.  This PR deliberately retained compatibility with Python 2.7, but the later MRs I referenced do transition to Python 3 completely.
- Dropping the embedded Genshi dependency might be possible, but I don't think it's achievable at present.  I'll need to investigate further, but when I looked recently for the Python 3.8 support, you had to run `2to3` over the Genshi code to make it run with Python 3.  It's for this reason I retained the embedded copy.  Maybe the copy that you get from pip has this done automatically.  I'll check.

I would suggest tackling these in follow-up PRs.  It will make these changes easier to review, and you'll also be able to look at the generated source changes to verify that they don't introduce any other changes following the odict change.

Regards,
Roger

sbesson:
Thanks. I agree getting this merged creates an useful intermediate state allowing to review changes without the `OrderedDict` changes.

In terms of dependencies, I was thinking about using `future` rather than `six` as this has been the dependency used everywher else. Dropping Python 2 support will undoubtedly happen at some point although we would like to maintain Py2/Py3 compatibility for at least the upcoming minor series.

Thanks again for this impressive contribution and let us know if you have some capacity to look into consuming the upstream Genshi dependency in the upcoming week.

PR #68, opened 2018-01-17
reword value domain constraint error messages

mtbc:
This PR simply changes the wording of some numerical exceptions' messages. My intent is to make the meaning clearer at first glance. When I first read https://www.openmicroscopy.org/qa2/qa/feedback/19174/ I was kind of nonplussed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+5.6.2 (2018-01-25)
+------------------
+
+* Report numerical value domain errors more clearly
+  [#68](https://github.com/ome/ome-model/pull/68)
+
 5.6.1 (2018-01-17)
 ------------------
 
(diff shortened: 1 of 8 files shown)


rleigh-codelibre:
I checked the C++ equivalent in `src/main/cpp/ome/xml/model/primitives/ConstrainedNumeric.h` but it's limited to `‘%1%’ not a supported value of %2%` as a catch-all.

sbesson:
Thanks @mtbc. Looking at the exception from the description of https://github.com/openmicroscopy/bioformats/pull/3021, should we also review the wording of the `NonNegativeInteger` exceptions?

sbesson:
Thanks. The new wording makes sense and should be suitable for a patch release of `ome-model`. What is your feeling with regard to the testing of these changes prior to bumping the version in the downstream components?

mtbc:
No strong opinions. The changes are,

* low risk
* easy to test (just try constructing some instances)
* of no urgency.

sbesson:
Agreed, do you want to push and extra commit to update `NEWS.md` with a summary of the changes? I will try to have a go at cutting the release today or tomorrow.

mtbc:
I don't see a `NEWS.md`?

sbesson:
Try to merge `origin/master` into your branch ?

PR #69, opened 2018-02-03
build: Unify component version property names

rleigh-codelibre:
This is to allow `-Dome-common.version=...` or rewriting the configuration to work the same for all components.  This one was different for some reason.

Testing: check builds are green.

### DIFF
--- a/ome-xml/pom.xml
+++ b/ome-xml/pom.xml
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>org.openmicroscopy</groupId>
       <artifactId>ome-common</artifactId>
-      <version>${ome_common.version}</version>
+      <version>${ome-common.version}</version>
     </dependency>
 
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -116,7 +116,7 @@
     <logback.version>1.1.1</logback.version>
     <slf4j.version>1.7.6</slf4j.version>
     <testng.version>6.8</testng.version>
-    <ome_common.version>5.3.1</ome_common.version>
+    <ome-common.version>5.3.1</ome-common.version>
 
     <ome.model.schemaver>2016-06</ome.model.schemaver>
     <ome.model.schemapath>specification/src/main/resources/released-schema/${ome.model.schemaver}</ome.model.schemapath>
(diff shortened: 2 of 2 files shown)


rleigh-codelibre:
Green in https://web-proxy.openmicroscopy.org/east-ci/job/BIOFORMATS-merge/20/

rleigh-codelibre:
I don't think it matters that much, it's not something that an end user would have any need or reason to override; until now we haven't even had a reason to do this ourselves.  But if you think it's necessary, I can create a compatibility variable.

joshmoore:
@rleigh-codelibre : not knowing who's consumed this over the [last year](https://github.com/rleigh-codelibre/ome-model/commit/cb9a5f7a1eae837963194befef83367bbb023115) can I suggest we get the alias added here along with a deprecated comment and then merge this along with https://github.com/ome/ome-poi/pull/6 and https://github.com/ome/ome-codecs/pull/10?

In terms of the release, it might be a "nice-to-have" at least to show that someone else can still go through the steps. But with other upcoming changes, perhaps now isn't the right time?

joshmoore:
@rleigh-codelibre : do you have a commit coming for this?

rleigh-codelibre:
Done.

Do we have a policy for deprecation and removals in the build systems?

I don't really consider this "API" myself.  It's developer-only build scripts which by their nature can change, and without impact on the real API in the jars.  No one will encounter this unless building from source, and even then if they are deliberately tampering with the properties, in which case it requires some effort on their part to keep track of changes in the poms.  I hope we can drop this sooner rather than later.

sbesson:
No formal policy and I would also rank build system changes as having less impact than API changes in general e.g. here the addition of a build property would be fine as a patch release from my side. Re property deprecation, aliasing is probably the closest we have although there might be formal ways to do so.

Regarding developer stability, my understanding of this ongoing work is that we are effectively establishing a stable scheme for property names across our set of Maven components. Effectively, we are not expecting subsequent changes to these property names outside the context of other major changes like component renaming or build system overhaul.

joshmoore:
Thanks both. I'll leave you guys to capture when/how we want to remove the compatibility layer.

Merging as per https://github.com/ome/ome-codecs/pull/10#issuecomment-370461418 for a patch bump.

Issue #71, opened 2018-03-07
Subcomponents configuration


joshmoore:
```
[ERROR] Rule failure while trying to close staging repository with ID "orgopenmicroscopy-1047".
[ERROR]
[ERROR] Nexus Staging Rules Failure Report
[ERROR] ==================================
[ERROR]
[ERROR] Repository "orgopenmicroscopy-1047" failures
[ERROR]   Rule "javadoc-staging" failures
[ERROR]     * Missing: no javadoc jar found in folder '/org/openmicroscopy/specification/5.6.3'
[ERROR]     * Missing: no javadoc jar found in folder '/org/openmicroscopy/ome-xml/5.6.3'
[ERROR]
[ERROR]
[ERROR] Cleaning up local stage directory after a Rule failure during close of staging repositories: [orgopenmicroscopy-1047]
[ERROR]  * Deleting context 226f0ceed4c8d6.properties
[ERROR] Cleaning up remote stage repositories after a Rule failure during close of staging repositories: [orgopenmicroscopy-1047]
[ERROR]  * Dropping failed staging repository with ID "orgopenmicroscopy-1047" (Rule failure during close of staging repositories: [orgopenmicroscopy-1047]).
```

similar to https://github.com/ome/ome-stubs/pull/4

rleigh-codelibre:
Should be fixed with the linked PR.

PR #72, opened 2018-03-14
build: Use plugins from root pom

rleigh-codelibre:
Close: #71 

Same idea as https://github.com/ome/ome-stubs/pull/4

Note: Also check that you have `JAVA_HOME` set since for some reason this is a requirement for javadoc generation.

Testing: Check builds are green, and check you can release.

/cc @joshmoore

### DIFF
--- a/ome-xml/pom.xml
+++ b/ome-xml/pom.xml
@@ -165,18 +165,54 @@
         </configuration>
       </plugin>
       <plugin>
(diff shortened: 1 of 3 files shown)


joshmoore:
Green, merging for a test run with the nexus plugins. Thanks, @rleigh-codelibre

PR #77, opened 2018-04-23
AffineTransform: Add createRotationTransform static method

rleigh-codelibre:
Move `[SubResolution]FormatReader.getRotationTransform` to the `AffineTransform` class itself.  Use a static method for Java, and a regular constructor for C++ (where it can be specialised on a radian unit).

Testing: Check builds remain green for both Java and C++.

This will permit dropping of this implementation from `FormatReader` once released, and complete removal from `SubResolutionFormatReader`.

### DIFF
--- a/xsd-fu/templates-cpp/OMEXMLModelObject.template
+++ b/xsd-fu/templates-cpp/OMEXMLModelObject.template
@@ -55,6 +55,10 @@
 
 #include <ome/common/log.h>
 
+{% if klass.name == "AffineTransform" %}\
+#include <ome/common/units/angle.h>
+
+{% end class is AffineTransform %}\
 #include <ome/common/xml/dom/Document.h>
 #include <ome/common/xml/dom/Element.h>
 #include <ome/common/xml/dom/Node.h>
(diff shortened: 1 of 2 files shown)


dgault:
Builds are green and the generated source code looks good.

Moving the implementation out of FormatReader to AffineTransform makes sense and doesn't cause any issues. Normally Id be reluctant about adding further exceptions to the templates but AffineTransform already has existing exceptions in both so it is fine in this case. 

My initial reaction looking at the model code would have been to use a constructor for Java also, but looking at the existing usage of AffineTransform I can see the static method making sense too. Are there any specific advantages in this particular case for using one over the other?

rleigh-codelibre:
Java doesn't have templated unit types, which means a direct use of a constructor in Java would be overloading on `double`.  C++, in constrast, is overloading on `angle-in-radians`, with transparent conversions from angles in other unit systems.  This lack of specificity in Java is why I left it as a static method; if we ever add methods for other transforms such as scaling, the `double` overload would be ambiguous.

dgault:
I could well see other transforms being added in the future so that makes sense to me. In that case Im happy for this to be merged.

rleigh-codelibre:
This is a trivial addition, so can go into the next minor release.

PR #81, opened 2018-06-23
Link fixes

rleigh-codelibre:
Closes: https://github.com/ome/ome-files-cpp/issues/98

This corrects some long-standing issues in both the C++ and Java reference processing implementations.

- `OMEModelObject::link` implementations always chained up to their parent before handling the linking themselves.  This works, but would result in the parent issuing a warning if it couldn't handle it, despite the child later handling the link.  We now chain up only if the child can't handle the link, and issue a warning right at the top if nothing handled it.  This solves a number of annoying Java model serialisation warnings which are completely bogus.  The same applies for C++.
- C++ `OMEXMLMetadataStore` `addReference` calls in `setXXXAnnotationRef` methods updated to match the Java templates more closely.  There are no functional changes.  A shared pointer copy is replaced by a cast.

In effect, you will see that setting an `AnnotationRef` on `Detector` or any other element derived from `ManufacturerSpec` will no longer have `ManufacturerSpec::link` issue a bogus warning.  The ome-files-cpp `metadata-formatwriter` example will no longer issue spurious warnings.  This applies to any model object which is derived from another model object, e.g. LightSource, where the reference is handled by the most derived type, or any derived type after the first, where the chaining up will trigger the spurious warning.

For OMERO import, this should remove a large number of the bogus import warnings related to OME-XML metadata which we have seen on e.g. the QA system and other channels over the last few years.  `LightSource` and `Shape` are notable culprits.

Testing:

- Check the C++ build logs; they will no longer contain `[warning] Unable to handle reference of type` messages for any of the unit tests.
- Run `ome-files` `examples/metadatastore-formatwriter test.ome.tiff` and then `tiffinfo` or `tiffcomment` on `test.ome.tiff`.  Check that you get three `AnnotationRef` elements linking to the three annotations, demonstrating that the reference processing and subsequent serialisation worked correctly.
- Run `ome-files info test.ome.tiff --omexml` to check this round-trips.
- Check Java with a similar sample for `metadata-formatwriter`, creating an `AnnotationRef` on `Detector` or similar.  You'll see the warnings go when built against this ome-model snapshot, but otherwise no other functional changes.

/cc @joshmoore

### DIFF
--- a/ome-xml/src/main/cpp/ome/xml/model/detail/OMEModelObject.cpp
+++ b/ome-xml/src/main/cpp/ome/xml/model/detail/OMEModelObject.cpp
@@ -38,6 +38,7 @@
 
 #include <ome/common/xml/String.h>
 
+#include <ome/xml/model/Reference.h>
 #include <ome/xml/model/detail/OMEModelObject.h>
 
 namespace ome
@@ -104,9 +105,14 @@ namespace ome
         }
 
         bool
-        OMEModelObject::link (std::shared_ptr<Reference>&                         /* reference */,
-                              std::shared_ptr<::ome::xml::model::OMEModelObject>& /* object */)
+        OMEModelObject::link (std::shared_ptr<Reference>&                         reference,
+                              std::shared_ptr<::ome::xml::model::OMEModelObject>& object)
         {
+          BOOST_LOG_SEV(logger, ome::logging::trivial::warning)
+            << elementName()
+            << " unable to handle reference of type "
+            << reference->elementName()
+            << " for " << object->elementName();
           return false;
         }
 
--- a/ome-xml/src/main/java/ome/xml/model/AbstractOMEModelObject.java
+++ b/ome-xml/src/main/java/ome/xml/model/AbstractOMEModelObject.java
@@ -44,11 +44,18 @@ import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * @author callan
  *
  */
 public abstract class AbstractOMEModelObject implements OMEModelObject {
+
+  /** Logger for this class. */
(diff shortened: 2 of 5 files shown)


rleigh-codelibre:
Copied to [GitLab](https://gitlab.com/codelibre/ome-model/merge_requests/91).

Generated source changes are in the [pipeline](https://gitlab.com/rleigh/ome-model/pipelines/30115273).

rleigh-codelibre:
See also [review note](https://gitlab.com/codelibre/ome-model/merge_requests/101) on GitLab.

rleigh-codelibre:
@joshmoore @mtbc Is there anything blocking this reference processing fix being merged?  It's a pretty straightforward fix.

mtbc:
:man_shrugging: So this can supersede #87? :+1:

rleigh-codelibre:
@mtbc No, they solve separate and unrelated issues.  This one prevents warnings being issued when there was no problem, i.e. they are false positives.  The other one is actually allowing the storage of invalid data in the model.  This one is a trivial change to where the warning is issued.  By always chaining up to the parent when the link wasn't handled, we warn only if the root object link method is entered, indicating that no subclass was capable of handling the situation.  Thus we only warn about legitimate cases.

snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#513](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/513/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/513/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - ome-xml/src/main/cpp/ome/xml/model/detail/OMEModelObject.cpp
    - xsd-fu/templates-cpp/OMEXMLMetadata.template
    - xsd-fu/templates-cpp/OMEXMLModelObject.template

--conflicts

sbesson:
Closing as this conflicts with the removal of the C++ classes in #111

PR #84, opened 2018-09-18
maven: Correct filtering of model version

rleigh-codelibre:
I'm not sure when this regression occurred.  It might date back to the repository decoupling.  Whenever it happened, this is the simple fix to enable resource filtering and package up the filtered resource with the correct model version.  Resource moved to `src/main/resources` where maven expects it by default.

Testing:

Without this PR:

```
% mvn
% mkdir check
% cd check
% jar xf ../ome-xml/target/ome-xml-5.6.4-SNAPSHOT.jar
% cat java/ome/xml/model/omemodel.properties
schema.version = ${ome.model.schemaver}
```

With this PR merged:

```
% mvn
% mkdir check
% cd check
% jar xf ../ome-xml/target/ome-xml-5.6.4-SNAPSHOT.jar
% cat ome/xml/model/omemodel.properties
schema.version = 2016-06
```


rleigh-codelibre:
I'm not sure there are any current users of the `schema.version` property, though there might have been in the past in bioformats?  I thought this was the means by which bioformats would query the model version in use, but maybe we never got around to enabling that after adding the property?  It's effectively public API which other people could conceivably be using though.  Clearly if no one has reported it before, it's not likely to be used in any significant capacity, but the fix is pretty trivial to make it work as intended.

snoopycrimecop:
Conflicting PR. Removed from build [OME-FILES-CPP-DEV-merge-push-superbuild#1057](https://ci.openmicroscopy.org/job/OME-FILES-CPP-DEV-merge-push-superbuild/1057/). See the [console output](https://ci.openmicroscopy.org/job/OME-FILES-CPP-DEV-merge-push-superbuild/1057/consoleText) for more details.
Possible conflicts:
  - PR #65 rleigh-codelibre 'Add support for building with Python 3 as well as Python 2'
    - requirements.txt

--conflicts

sbesson:
I have been looking for the history of this property. It looks like it was introduced in https://github.com/qidane/bioformats/pull/7 and implemented in `ome.xml.model.MapPairs` which was removed in https://github.com/openmicroscopy/bioformats/pull/2427. Apart from this single usage, it looks like the schema version is hardcoded in our stack although with this change, it could conceivably be read dynamically from the properties file.

The proposed change looks straightforward. I will exclude #65 which conflicts with this PR and that no-one has had the capacity to review so that this can get validated by the daily CI builds and merged.

rleigh-codelibre:
I'll correct #65 once this is merged.

PR #87, opened 2018-10-08
ignore null references rather than noting them for resolution

mtbc:
`convertMetadata` seems to copy `null` property values then when resolving references we get lots of,

> ... reference to null missing from object hierarchy.

For instance, see https://www.openmicroscopy.org/qa2/qa/feedback/21779/ or a server log from [OMERO-DEV-merge-integration](https://ci.openmicroscopy.org/job/OMERO-DEV-merge-integration/).

I suspect that these null referents are valueless and should be ignored in the first place so this PR adjusts `addReference` accordingly. If we really want to track them then perhaps we could instead adjust `resolveReferences()` to not WARN about them or change `MetadataConverter` to not set them.

I do not see where the return value of `addReference` is documented so I guessed.

Really it looks as if `references` should be a [SetMultimap](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/SetMultimap.html) but that is outwith this PR.

### DIFF
--- a/ome-xml/src/main/java/ome/xml/model/OMEModelImpl.java
+++ b/ome-xml/src/main/java/ome/xml/model/OMEModelImpl.java
@@ -99,6 +99,9 @@ public class OMEModelImpl implements OMEModel {
    */
   @Override
   public boolean addReference(OMEModelObject a, Reference b) {
+    if (b != null && b.getID() == null) {
+      return false;
+    }
     List<Reference> bList = references.get(a);
     if (bList == null) {
       bList = new ArrayList<Reference>();


rleigh-codelibre:
A deeper question is why the source metadata store has null values to begin with, which is likely a serious bug in and of itself.  That should not be possible!  What is creating these null values?

mtbc:
Next time I get some relevant client code to a working state, I'll plug a debugger back in, remind myself of the details and report back.

mtbc:
So, in one example:

1. `ome.xml.meta.MetadataConverter.convertImages` uses `getImageExperimentRef`.
1. The `experimentRef` string is `null`.
1. `MetadataConverter.convertImages` now does `setImageExperimentRef` with that `null` value.
1. `OMEXMLMetadataImpl.setImageExperimentRef` packs that `null` into the ID of an `ExperimenterRef` and adds a reference from the image to it.
1. Later on we hit `ome.xml.model.OMEModelImpl.resolveReferences()`.
1. We loop through the `Reference` instances and attempt `getModelObject` on their IDs.
1. We WARN about the `null` response from `getModelObject` to the `null` ID we earlier set.

mtbc:
Ah, given my above description I should update my PR, sorry.

mtbc:
Of course, could instead bump such logging down to DEBUG or INFO if we really do want to see this stuff for some reason.

rleigh-codelibre:
@mtbc Logging would be a good idea; we already do this for other cases of inconsistency.  Examples are [here](https://github.com/ome/ome-model/blob/master/ome-xml/src/main/java/ome/xml/model/OMEModelImpl.java#L142) and [here](https://github.com/ome/ome-model/blob/master/xsd-fu/templates-java/OMEXMLModelObject.template#L496).  The latter should *not* be debug class, it should be warn; it's only due to https://github.com/ome/ome-model/pull/81 not being fixed which prevents it being a regular warning.  (That PR should probably make it a warning as well.)

mtbc:
I was thinking that just the `Reference.getID() == null` case (i.e., that which this PR has us ignore) could be less than WARN seeing as it comes up so commonly in normal workflows; happy to leave other cases unaffected.

rleigh-codelibre:
Ideally, it should be warn.  If we could enable logging with a stacktrace and pinpoint where the erroneous nulls are being introduced and fix them, it would cease to be a common occurence and be a useful warning.  Setting to debug like the other one would probably be better in the interim.

mtbc:
Re. https://github.com/ome/ome-model/pull/87#issuecomment-429341606 where do you think the actual fault is: being able to get the `null`s, the setting them, or constructing and noting `Reference` instances for  them?

(Though for https://trello.com/c/XXw8Tbf5/21-resolving-null-references if there is anything non-trivial to be done then that will need separately prioritizing and scheduling: for my current work I'd just find some different way to stop these annoying messages filling the command-line user's console. People's Blitz logs can wait for another time.)

rleigh-codelibre:
@mtbc For the metadatastore, it should not be possible to return null here.  For example, this is the C++ implementation:

```
      const std::string&
      OMEXMLMetadata::getImageExperimentRef(index_type imageIndex) const
      {
        // ExperimentRef is reference and occurs only once
        std::shared_ptr<ome::xml::model::Experiment> annotation(safe_fetch(safe_fetch(root, "getImageExperimentRef [root]")->getImage(imageIndex), "getImageExperimentRef [getImage(imageIndex)]")->getLinkedExperiment().lock());
        if (annotation)
          return annotation->getID();
        throw MetadataException("OMEXMLMetadata", "getImageExperimentRef",
                                "Internal metadata store inconsistency: null annotation");
      }
```

Note that it (1) returns a const reference; it's not possible to be null and that (2) it throws if it's not possible to return a value.  The underlying model function is:

```
        const std::weak_ptr<ome::xml::model::Experiment>
        getLinkedExperiment () const;
```

Which can be null.  However, it's part of the metadata store design that it throws exceptions rather than null pointers.

The Java implementation is:

```
  public String getImageExperimentRef(int imageIndex)
  {
    // ExperimentRef is reference and occurs only once
    return root.getImage(imageIndex).getLinkedExperiment().getID();
  }
```

and

```
  // Reference
  public Experiment getLinkedExperiment()
  {
    return experiment;
  }
```

The Java implementation should be the same.  If `experiment` is null, you will get an NPE.  If it's not null, it will be able to return the experiement ID, which should never, ever be null.

The conversion code:

```
              transfer(&MR::getImageExperimentRef,            &MS::setImageExperimentRef,            i);
```
or
```
      try {
        String experimentRef = src.getImageExperimentRef(i);
        dest.setImageExperimentRef(experimentRef, i);
      }
      catch (NullPointerException e) { }
```

The C++ code does the same try/catch inside the templated transfer function.  It should have thrown in the Java case, but did not.  This implies an inconsistency: the experiment exists but has no ID set.  This is invalid OME-XML, and should not be happening.

Hope that explains things,
Roger

mtbc:
> The Java implementation should be the same. If `experiment` is null, you will get an NPE. 

Okay, I fired up the debugger again. `getImageExperimentRef` returns `null` because it's `DummyMetadata`'s implementation: my `ome-model/ome-xml/target/generated-sources/xsd-fu/ome/xml/meta/DummyMetadata.java` includes:

```java
  public String getImageExperimentRef(int imageIndex)
  {
    return null;
  }
```

I'd guess that these should be throwing instead then?

rleigh-codelibre:
Ahh, makes sense.  I think every `null` return here should all be throwing `NullPointerException` to match the OMEXMLMetadataImpl implementation.  That's quite a huge behaviour discrepancy, and it does explain the misbehaviour nicely.  The -1 return values for integers could also be suspicious; most of these should also throw due to the parent objects no existing conceptually.

To be honest, I'm not sure why DummyMetadata isn't a subclass of OMEXMLMetadataImpl with a null or empty root, so that it behaves exactly the same but with zero data being stored.  For example:

```
  public int getReagentAnnotationRefCount(int screenIndex, int reagentIndex)
  {
    return -1;
  }
```

If the screen or reagent don't exist, this should throw getting the screen.  So should also throw by default.

It really depends upon what the documented design is.  Maybe null and -1 returns are OK, and all callers should be prepared to deal with this.  Maybe the `OMEXMLMetadataImpl` implementation is throwing when it should be returning null?  Either way, the existing implementations have two completely different mechanisms for propagating errors, and this places a burden on all callers to cater for both eventualities.  `MetadataConverter` is not at all set up to deal with null returns, which makes me think that `OMEXMLMetadataImpl` is correct.  It certainly made sense for the C++ implementation where objects can't be null when passed by reference.  But it does make running a debugger more complex due to it throwing hundreds of times by default when doing a conversion (so, not really exceptional).  However, it does make the API harder to misuse, even if it is not the nicest design.

mtbc:
I am also finding that `OMEXMLMetadataImpl` is such that `getBooleanAnnotationCount()` can throw `NullPointerException` because it doesn't have the null-check for `OME.getStructuredAnnotations()` that, say, `setBooleanAnnotationID` does. I had been too optimistic in expecting a zero count.

rleigh-codelibre:
Yes, I've always disliked this behaviour, it makes things more complicated and fragile for zero benefit.  I'm not sure why intermediate containers like `StructuresAnnotations` couldn't be created unconditionally and assumed to exist.  They could be elided on serialisation if empty.

rleigh-codelibre:
I also noticed [this comment](https://github.com/ome/ome-model/blob/master/xsd-fu/templates-cpp/AggregateMetadata.template#L27) in the `AggregateMetadata` template:

```
/// TODO: No handling of -1 as for Java. Callee must throw.
```

which shows it was something we had vaguely on the radar for a while.

mtbc:
Better if I replace this PR with a card referencing this discussion?

rleigh-codelibre:
@mtbc Did any discussion happen on a card in the interim?  It's likely I'll be looking at this on the C++ side soon, updating [this old MR](https://gitlab.com/codelibre/ome/ome-model/merge_requests/116), and also updating the Java code to match, so would be useful to know if anything is planned or already done.  I'll most likely leave the existing NULL checks in place on addition, and update the metadatastore implementations to provide the behaviour discussed above.

Thanks,
Roger

mtbc:
I don't recall anything outwith our discussion on this PR though I don't attend most of the Bio-Formats meetings. I suspect that, rather than the change I suggest here, the class broadly needs to be made less surprising or more helpful. In the meantime on the caller side I just use the blunt instrument of `DebugTools.enableLogging("ERROR");` to avoid the useless clutter.

mtbc:
https://github.com/ome/ome-model/pull/87#issuecomment-429883516 and https://github.com/ome/ome-model/pull/87#issuecomment-431372851 point to deeper design issues that would benefit from reconsideration in some breaking overhaul of Bio-Formats' API. In the meantime this PR does nothing more useful than paper over what may be a very deep hole.

PR #88, opened 2018-10-09
MetadataConverter: refactor Channel copying into separate public methods

melissalinkert:
Backported from unmerged private PR.  Adds two public ```convertChannelMetadata``` methods to allow just one ```Channel``` and all of its linkages to be copied from a ```MetadataStore``` to a ```MetadataRetrieve```.  The behavior of ```convertMetadata(MetadataRetrieve, MetadataStore)``` should be unchanged.

This will allow ```loci.formats.ChannelSeparator``` to copy all channel metadata when performing channel separation, instead of just channel names.  See https://github.com/openmicroscopy/bioformats/blob/develop/components/formats-bsd/src/loci/formats/ChannelSeparator.java#L325

### DIFF
--- a/ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
+++ b/ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
@@ -106,6 +106,267 @@ public final class MetadataConverter {
     convertRootAttributes(src, dest);
   }
 
+  /**
+   * Copy all metadata from the specified channel in the source MetadataRetrieve
+   * to the specified channel in the destination MetadataStore.  The specified
+   * destination Image must exist (e.g. via calling
+   * <code>setImageID(..., destImage)</code>), but the destination Channel only
+   * needs to exist if the copyID flag is <code>false<code>
+   * LightSource references are assumed to be correct and are not checked for validity.
+   *
+   * @param src the source {@link MetadataRetrieve}
+   * @param srcImage the index of the Channel's parent Image in src
+   * @param srcChannel the index of the Channel in src
+   * @param dest the destination {@link MetadataStore}
+   * @param destImage the index of the Channel's parent Image in dest
+   * @param destChannel the index of the Channel in dest
+   * @param copyID false if the Channel ID should be omitted from the copy operation
+   */
+  public static void convertChannelMetadata(MetadataRetrieve src, int srcImage, int srcChannel,
+    MetadataStore dest, int destImage, int destChannel, boolean copyID)
+  {
+    convertChannelMetadata(src, srcImage, srcChannel, dest, destImage, destChannel, copyID, null);
+  }
+
+
+  /**
+   * Copy all metadata from the specified channel in the source MetadataRetrieve
+   * to the specified channel in the destination MetadataStore.  The specified
+   * destination Image must exist (e.g. via calling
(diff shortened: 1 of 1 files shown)


ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
```diff
@@ -106,6 +106,267 @@ public static void convertMetadata(MetadataRetrieve src, MetadataStore dest) {
...
+   *
+   * @param src the source {@link MetadataRetrieve}
+   * @param srcImage the index of the Channel's parent Image in src
+   * @param srcChannel the index of the Channel in src
+   * @param dest the destination {@link MetadataStore}
+   * @param destImage the index of the Channel's parent Image in dest
+   * @param destChannel the index of the Channel in dest
+   * @param copyID false if the Channel ID should be omitted from the copy operation
+   */
+  public static void convertChannelMetadata(MetadataRetrieve src, int srcImage, int srcChannel,
```
sbesson: All other methods of this class are called `convert<Objects>`. Should this become `convertChannels`?

ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
```diff
@@ -106,6 +106,267 @@ public static void convertMetadata(MetadataRetrieve src, MetadataStore dest) {
...
+    MetadataStore dest, int destImage, int destChannel, boolean copyID)
```
sbesson: Most `convert` methods act on `(MetadataRetrieve, MetadaStore)` arguments with the exception of `convertLightSources(MetadataRetrieve, MetadataStore, int, List<String>)`. 

Two questions:
- do we want to unify the argument ordering?
- is there a concrete use case for have `srcImage` and `destImage` rather than `imageIndex` (https://github.com/openmicroscopy/bioformats/pull/3243/files has the same image index but different channel indices).

ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
melissalinkert: Done in abf622e.

ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
melissalinkert: Should the new argument ordering then be ```MetadataRetrieve src, MetadataRetrieve dest, int srcImage, int srcChannel, int destImage, int destChannel, boolean copyID)```?

I can imagine a case for needing both ```srcImage``` and ```destImage```.  For HCS datasets with one plane per TIFF/STK/etc., copying all of the data from Image 0 Channel 0 in the helper reader's  ```MetadataRetrieve``` to some other Image/Channel in the HCS reader's ```MetadataStore``` could be useful.  If a simpler signature with only ```imageIndex``` would help, though, it could easily be added and just delegate appropriately.

sbesson:
Merging for inclusion in a milestone release of `ome-model:6` and testing in Bio-Formats. I have not strong opinion on the best signature for https://github.com/ome/ome-model/pull/88#discussion_r247547262, this can be briefly discussed alongside other API points to review for Bio-Formats 6.

sbesson:
Now deployed as `org.openmicroscopy:ome-xml:6.0.0-m1`

PR #91, opened 2018-11-14
Add experimental OME-XML generation

joshmoore:
Combining work on/for:

 * https://github.com/IDR/idr0047-neuert-yeastmRNA/pull/1
 * https://github.com/spacetx/starfish/issues/603

this adds a simple Python interface (ome_model.experimental)
that will be deployed to PyPI for building OME-XML companion
files. Longer-term: these classes and methods should be code-
generated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2018 University of Dundee.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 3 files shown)


joshmoore:
This will need gh-90 to go green.

setup.py
```diff
@@ -0,0 +1,53 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+import os
+import sys
+
+from setuptools import setup
+
+
+version = '5.6.4.dev1'
```
sbesson: As mentioned separately, my only question is whether we should call this `6.0.0.dev1`

sbesson:
Merging and I will open a follow-up change allowing to cover synthetic SPW samples - see http://lists.openmicroscopy.org.uk/pipermail/ome-users/2018-November/007281.html

PR #92, opened 2018-11-22
Expand experimental model library to generate SPW samples

sbesson:
See http://lists.openmicroscopy.org.uk/pipermail/ome-users/2018-November/007281.html

This builds on top of #91 to create companion OME-XML for SPW use cases.

### DIFF
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # Generate companion files
 
-import os
 import re
 import sys
 import uuid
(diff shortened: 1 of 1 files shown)


joshmoore:
Can it still create image XML? Perhaps this should be `plate.companion()` and `image.companion()`.

sbesson:
:) I was debating between `create_{plate,image}_companion` or `create_companion(plate=plate, images=images)` but I like the idea of `Object.companion()`. In that case, would `plate.companion()` traverse the whole graph or does it stop at `ImageRef` and individual `image.companion` need to be invoked?

joshmoore:
The latter sounds complicated. I'm fine with either of the first two signatures to start with, but I definitely need to be able to generate non SPW images.

joshmoore:
This has been raised again by the SpaceTx crew. Do you want to get this into the first release? or shall I push the previous version to PyPI?

sbesson:
If that is useful for others to consume, I can definitely see getting this integrated and pushed to PyPi. For 6 pre-releases, this is is where the [semver](https://semver.org/#spec-item-1) vs [PEP440](https://www.python.org/dev/peps/pep-0440/) incompatiblity might be tricky to handle - maybe `6.0.0a1` ?

joshmoore:
Merging and looking into publishing.

joshmoore:
https://pypi.org/project/ome-model/6.0.0a1/ -- note: alpha version downloads without `--pre` as the only version.

Issue #89, opened 2018-10-29
Specification: changes to include OWL ontologies


sbesson:
## Introduction

Following the presentation of the Riken Metadatabase at the [2018 OME Users meeting](http://downloads.openmicroscopy.org/presentations/2018/Users-Meeting/Lightning-Talks/Norio_Kobayashi.pdf), @norikoba has been driving the translation of the [OME xsd schema](http://www.openmicroscopy.org/Schemas/OME/2016-06/ome.xsd) into OWL ontology. The current version is available as an incubator project at https://gitlab.com/openmicroscopy/incubator/ome-owl/. As discussed recently during the OME team visit in Kobe, this work is now reaching production level. 

The ome-model repository was created and consumed starting from Bio-Formats 5.2.0 and OME-Files 0.3.0. It is composed of two components:

- the `specification` component contains the XSD schemas, the XSLT transforms and the OME-XML samples for all releases of the OME data model as well as some utility Java classes for creating objects
- the `ome-xml` component contains Java and C++ code-generated classes for working with the model

This issue discusses how to integrate of the OWL ontologies as part of the `specification` component. None of the changes below are formally breaking the specification itself or any of the API although the new file layout would require a few changes discussed in each section.

## Specifications

Each release of the data model is versioned using a calendar format of type `YYYY-MM` and the corresponding XSD schemas are stored under a `YYYY-MM` folder under [specification/src/main/resources/released-schema/](https://github.com/ome/ome-model/tree/v5.6.3/specification/src/main/resources/released-schema).

It was agreed that the OWL ontology should use the same versioning strategy `YYYY-MM` for consistency. We could simply introduce another top-level folder under `specification/src/main/resources/`.  While doing so, it might be worth cleaning up the `released-` suffix.

### Proposal

- rename the XSD schemas folder as `specification/src/main/resources/schemas/YYYY-MM/`
- update all code in `specification`, OME-XML Java and C++ classes  as well as downstream components classes to consume the new schemas location
- create `specification/src/main/resources/ontologies/2016-06/` and move the 2016-06 ontology alongside the XSD schema

## Transforms

XSLT transforms are currently located under `specification/src/main/resources/transforms/`. The way `ttl` files will be transformed between different versions of the ontology is still not clear to me. Discussions seemed to indicate a process consuming the new ontology as well as the inference engine. No action needed at the moment although transform files could be added to this folder if necessary.

## Samples

[...]

PR #93, opened 2018-12-09
OME ontologies: schemas and samples reorganization

sbesson:
Implements the first part of the proposal described in #89 

- `specification/samples` are migrated under `specification/src/main/resources/samples` and should be packaged as part of the specification JAR
- the XSD schemas under`specification/src/main/resources/relased-schemas` are renamed as `specification/src/main/resources/relased-schemas`
- the corresponding references across the repository (publication script, Cmake files) are adjusted to consume the new paths

With these changes, it should be possible to create `specification/src/main/resources/ontologies` and add additional samples to `specification/src/main/resources/samples`.

/cc @joshmoore @norikoba


sbesson:
After second thought, closing this PR.

Although I still believe we want to fix the name eventually, renaming the `released-schemas/` directory might just put unnecessary last minute changes while we try to review and freeze the various components for Bio-Formats 6. More importantly, it does not preclude the addition of the 2016-06 ontology as another top-level directory.

Similarly, packaging the samples as part of the JAR (3977bc7) is hopefully valuable but it should be  good to drive this feature using a concrete use case.

PR #105, opened 2020-01-05
Python3 cleanup and add Python 3.8 support

rleigh-codelibre:
This PR is a followup for #65, containing the followup work I mentioned there.

- Update code to be Python 3-only
- Update Genshi to latest upstream, a requirement for Python 3.8
- Running Genshi with Python 3.8 requires 2to3 running over it, making the result Python 3-only.  Genshi is not runnable on 2.7 and 3.8 from a single source; upstream runs 2to3 over it when building the wheel, same as done here
- Remove all remaining use of `six` and `__future__` except for in imported third-party Genshi code
- Clean up xsd-fu by removing obsolete and broken content, and tidy up the templates
- No generated source changes

This PR contains previous work from:
- [GitLab !143](https://gitlab.com/codelibre/ome/ome-model/merge_requests/143)
- [GitLab !144](https://gitlab.com/codelibre/ome/ome-model/merge_requests/144)
- [GitLab !150](https://gitlab.com/codelibre/ome/ome-model/merge_requests/150)

In [this comment](https://github.com/ome/ome-model/pull/65#issuecomment-552509508), @sbesson mentioned retaining Python 2.7 support.  Unfortunately, it seems to be the case that Python 2.7 and Python 3.8 are mutually exclusive when it comes to Genshi.  It looks like in the interim that Genshi 0.7.3 is now pip installable, which might mitigate that.  However, I don't have time to dedicate to Python 2, and Genshi has historically broken repeatedly when installed via pip for various reasons, which is why it's under version control in the first place.  Python2 is not supported since the [1st of January](https://www.python.org/doc/sunset-python-2/), by the way.  Since it's a build-time dependency only, it should not be at all onerous to drop it for this module.

Testing: Generated source code should be identical to current master


rleigh-codelibre:
Looks like Travis MAVEN_OPTS handling is broken, and has been for some time.  This only showed up when strictly requiring Python 3, because Travis seems to have been silently testing both the Python 2 and Python 3 variants with python2 *only*.

See [here](https://github.com/travis-ci/travis-ci/issues/4613#issuecomment-135215956) and [here](https://stackoverflow.com/questions/29201549/travis-ci-ignoring-maven-opts) for further information.  I've not yet found a clean workaround to pass options to maven, but I'll continue looking.

dgault:
Thanks @rleigh-codelibre for the continued effort, its a great help.

Testing the new generated sources and they remain the same as previous which is as expected. The updates to Genshi look good and everything is building and running as expected so I see no issue with the upgrade. 

The cleanup to xsd-fu likewise has no negative impact on anything we are currently using and seems safe to remove.

The Travis workaround at least has Travis back up and running with the last commit. Overall Im happy to get this work merged and any improvements to it can be handled in a follow up PR.

PR #106, opened 2020-01-16
OME-XML: Update python in pom to python3

dgault:
This is a supplementary PR to https://github.com/ome/ome-model/pull/105 and should fix the broken build as seen in https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-build/jdk=JDK8,label=testintegration/248/console

### DIFF
--- a/ome-xml/pom.xml
+++ b/ome-xml/pom.xml
@@ -79,7 +79,7 @@
 
   <properties>
     <project.rootdir>${basedir}/..</project.rootdir>
-    <python>python</python>
+    <python>python3</python>
     <xsdfu.script>../xsd-fu/xsd-fu</xsdfu.script>
     <xsdfu.modelpath>${basedir}/target/generated-sources/xsd-fu</xsdfu.modelpath>
     <xsdfu.metadatapath>${basedir}/target/generated-sources/xsd-fu</xsdfu.metadatapath>


rleigh-codelibre:
Hi David,

Just one point to note, that this isn't strictly portable.  The most robust solution would be to override the property on the command line.

While `python3` is often present, it's not universally true.  It's not true in virtualenvs on several platforms.  And it also won't be true after python2 is fully dropped and systems make `python` invoke python3 there as well.

So it's fine to swap, but it will break as much the other way around as it does the current way around, but that will become less likely over time.

Regards,
Roger

joshmoore:
e.g. https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-image/274/console

I've opened https://github.com/ome/bio-formats-build/pull/73 as a temporary fix. An alternative would be to take a parameter in the [Dockerfile itself](https://github.com/ome/bio-formats-build/blob/master/Dockerfile#L19) then Jenkins could pass through a variable.

rleigh-codelibre:
@dgault @joshmoore You might find this of interest:

[machines.yml](https://gitlab.com/codelibre/ome/ome-gitlab-ci/blob/master/machines.yml) -- look at the various `PYTHON_EXECUTABLE` variants; this demonstrates the variance between `python` and `python3` on various platforms.

We then run gradle with `"-Ppython=$PYTHON_EXECUTABLE"`; the equivalent can be done with maven.

sbesson:
--exclude see https://github.com/ome/bio-formats-build/pull/79

Issue #103, opened 2019-09-24
Add support for Creator attribute in ome.xml.meta classes


sbesson:
Initially reported by and discussed with @ngladitz over IRC

The generated OME model classes have currently only partial support for the [OME.Creator](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#OME_Creator) attribute of the OME 2016-06 schema. More precisely

- the low-level model classes have the correct getters and setters - see [ome.xml.model.OME.getCreator()](https://static.javadoc.io/org.openmicroscopy/ome-xml/6.0.1/ome/xml/model/OME.html#getCreator--) and [ome.xml.model.OME.getCreator(String)](https://static.javadoc.io/org.openmicroscopy/ome-xml/6.0.1/ome/xml/model/OME.html#setCreator-java.lang.String-)
- the metadata classes lack such getters/setters - see [ome.xml.meta.MetadataRetrieve](https://static.javadoc.io/org.openmicroscopy/ome-xml/6.0.1/ome/xml/meta/MetadataRetrieve.html)

From the Bio-Formats consumer perspective, this means that the `Creator` metadata is currently lost while running `MetadataConverter`.

Unless there is a clear rationale for excluding the `Creator` attribute, it feels like we want to (re)add support in the metadata classes. Assuming the top-level OME element is not completely handled by the code generation, a starting point would be to add methods to the list of manual definitions e.g. https://github.com/ome/ome-model/blob/v6.0.1/xsd-fu/templates-java/MetadataRetrieve.template#L348. As an additional technical point, it would be very useful to find a way and make such additions amenable to a minor release increment (backwards-compatible API addition) rather than a major release increment (breaking API).

sbesson:
Closing as per #108

PR #108, opened 2020-03-17
OME/@Creator

ngladitz:
This is meant to add broader support for the `OME/@Creator` attribute.
Specifically to fix issue #103 

I primarily tried to duplicate the existing logic I found for `OME/@UUID`.
I am not sure I understand every context where I applied these changes and I am not sure I found every relevant context.

A such I would appreciate an appropriately critical review of these changes.

### DIFF
--- a/xsd-fu/templates/cpp/MetadataRetrieve.template
+++ b/xsd-fu/templates/cpp/MetadataRetrieve.template
@@ -436,6 +436,14 @@ namespace ome
         getUUID() const = 0;
 
         /**
+         * Get the Creator associated with this collection of metadata.
+         *
+         * @returns the Creator.
+         */
+        virtual const std::string&
+        getCreator() const = 0;
+
+        /**
          * Get the map value of MapAnnotation.
          *
          * @param mapAnnotationIndex the MapAnnotation index.
--- a/xsd-fu/templates/cpp/MetadataRetrieve.template
+++ b/xsd-fu/templates/cpp/MetadataRetrieve.template
@@ -431,8 +431,6 @@ namespace ome
          * Get the UUID associated with this collection of metadata.
          *
          * @returns the UUID.
-         *
-         * @todo Can this be a const reference?
          */
         virtual const std::string&
         getUUID() const = 0;
--- a/xsd-fu/templates/cpp/MetadataStore.template
+++ b/xsd-fu/templates/cpp/MetadataStore.template
@@ -251,6 +251,14 @@ namespace ome
         setUUID(const std::string& uuid) = 0;
 
         /**
+         * Set the Creator associated with this collection of metadata.
+         *
+         * @param creator the Creator to set.
+         */
+        virtual void
+        setCreator(const std::string& creator) = 0;
+
+        /**
          * Set the map value of MapAnnotation.
          *
          * @param value the map value to set.
--- a/xsd-fu/templates/cpp/OMEXMLMetadata.template
+++ b/xsd-fu/templates/cpp/OMEXMLMetadata.template
@@ -1009,6 +1009,24 @@ ${counter(k, o, v)}\
 
 {% if fu.SOURCE_TYPE == "header" %}\
         // Documented in base class.
+        const std::string&
+        getCreator() const;
+{% end header %}\
+{% if fu.SOURCE_TYPE == "source" %}\
+      const std::string&
+      OMEXMLMetadata::getCreator() const
+      {
+        std::shared_ptr<const std::string> creator = root->getCreator();
+        if (creator)
+          return *creator;
+        else
+          throw MetadataException("OMEXMLMetadata", "getCreator",
+                                  "Internal metadata store inconsistency: null object");
+      }
+{% end source %}\
+
+{% if fu.SOURCE_TYPE == "header" %}\
+        // Documented in base class.
         const ome::xml::model::primitives::OrderedMultimap&
         getMapAnnotationValue(index_type mapAnnotationIndex) const;
 {% end header %}\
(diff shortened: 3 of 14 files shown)


rleigh-codelibre:
@ngladitz The changes all look reasonable, with one caveat.  Other than the MetadataConverter changes, all the rest of the MetadataStore/Retrieve changes should have been automatically code generated with xsd-fu without the need to hardcode changes in all of the templates.  I don't understand (without deeper digging) why the Creator attribute is special in this regard.  However, there are other hardcoded additions in the templates already, and it might be that there's an underlying cause which underlies this code generation defect.

None of the above is to suggest that the change should not be merged as is, since there's precedent for this type of change, but we might want to look at investigating and fixing some defects in xsd-fu.

rleigh-codelibre:
Regarding the C++ template and MetadataConverter changes, there are problems here.  Please see the [GitLab build pipeline](https://gitlab.com/rleigh/ome-model/pipelines/127187347) for more details.

xsd-fu/templates/cpp/MetadataRetrieve.template
```diff
@@ -431,12 +431,18 @@ namespace ome
          * Get the UUID associated with this collection of metadata.
          *
          * @returns the UUID.
-         *
-         * @todo Can this be a const reference?
-         */
+         /
```
rleigh-codelibre: Broken comment terminator.

xsd-fu/templates/cpp/MetadataRetrieve.template
```diff
@@ -431,12 +431,18 @@ namespace ome
...
         virtual const std::string&
         getUUID() const = 0;
 
+        /**
+         * Get the Creator associated with this collection of metadata.
+         *
+         * @returns the Creator.
+         /
```
rleigh-codelibre: Broken comment terminator.

xsd-fu/templates/cpp/OMEXMLMetadata.template
```diff
@@ -1158,6 +1176,20 @@ ${getter(k, o, prop, v)}\
       }
 {% end source %}\
 
+{% end debug %}\
```
rleigh-codelibre: Should be
```
{% if fu.SOURCE_TYPE == "header" %}\
```

xsd-fu/templates/cpp/OMEXMLMetadata.template
```diff
@@ -1007,6 +1007,24 @@ ${counter(k, o, v)}\
...
+        const std::string&
+        getCreator() const;
+{% end header %}\
+{% if fu.SOURCE_TYPE == "source" %}\
+      const std::string&
+      OMEXMLMetadata::getCreator() const
+      {
+        std::shared_ptr<const std::string> creator = root->getCreator();
+        if (creator)
+          return *root->getCreator();
```
rleigh-codelibre: Would be nicer as `return *creator;`, likewise for `getUUID()`.

rleigh-codelibre:
@ngladitz In addition to the comments above, please see [the last two commits on GitLab](https://gitlab.com/rleigh/ome-model/-/commits/ome-creator-gl) for the needed fixes, and [this GitLab pipeline](https://gitlab.com/rleigh/ome-model/pipelines/127197072) for the passing builds and generated source changes diffs for both C++ and Java.  Both look fine other than the minor pointer dereference I mentioned above for C++.

@sbesson Are there any plans to re-establish C++ CI testing by OME?  Without it, the chance of regressions is very high.  If you don't want to keep it around, I could open up a PR to drop it if that is preferable.

ngladitz:
@rleigh-codelibre thank you! I hope I fixed all the issues you found (or rather correctly incorporated your fixes).

rleigh-codelibre:
@ngladitz Thanks!  There's one missing change remaining: [OMEXMLMetadata conditional](https://gitlab.com/rleigh/ome-model/-/commit/b6af24a4e6639ab83cbf9621f0d3d94a29f94705), tested [in this pipeline](https://gitlab.com/rleigh/ome-model/pipelines/127298716).

ngladitz:
Ah sorry I hope I got them all this time; thanks again!

xsd-fu/templates/cpp/MetadataStore.template
```diff
@@ -250,6 +250,14 @@ namespace ome
         virtual void
         setUUID(const std::string& uuid) = 0;
 
+        /**
+         * Set the Creator associated with this collection of metadata.
+         *
+         * @param creator the Creator to set.
+         */
+        virtual void
+        setCreator(const std::string& creator) = 0;
```
sbesson: The main issue with this change and `getCreator` below is that the addition of methods to the `MetadataRetrieve/MetadataStore` interface is a breaking change. As an example this broke downstream code implementing the `MetadataStore` interface.

Given the OME codebase now requires Java 8 and above, would it be possible to use default methods to workaround this issue and make this change a backwards-compatible API addition?

rleigh-codelibre: This should only need doing for the Java implementation.  There are no known users of the C++ implementation other than OME Files C++, and that's no longer consuming the version in this repository.

ngladitz:
@sbesson I've pushed two new commits; one that actually bumps the java version from 7 to 8 and one that adds the default functions to the java interfaces (and removes the now redundant implementation from `DummyMetadata`). Assuming this is what you expected I could squash the default method commit and put the version bump commit before it.

rleigh-codelibre:
@ngladitz Your latest changes pass without any problems: [GitLab pipeline](https://gitlab.com/rleigh/ome-model/pipelines/127513559).

The Java changes all look fine as well.

I do wonder with the change to Java 8 if we could (as a future improvement) create similar stubs for all MetadataStore and MetadataRetrieve interfaces, leaving DummyMetadata as an empty class extending that interface.

ngladitz:
Hi thanks again ... @sbesson I think I might have missed and due to force pushing lost the explicit changes you requested. Do the current changes encompass what you proposed?

sbesson:
Thanks @ngladitz, yes the default methods implemented in 848e76dee15397d9eb1c1dda15c8c1bee2452de6 correspond to what I had in mind. I can confirm they are sufficient to avoid implementations of the `MetadataStore` interface outside this component to break and make this change amenable for a minor release increment.

I spend some time trying to construct a minimal example exercising these new Java interfaces for example when converting a file in Bio-Formats. However, I still need to work around some of the downstream code which is overriding the `OME@Creator`. I will keep working towards an example validating the proposed change early next week so that we can be in a position to merge this PR.

sbesson:
Thank you @ngladitz. Merging after a final sign-off from the @ome/formats and we will work towards enabling this in the upcoming Bio-Formats minor release.

ngladitz:
Thanks!

PR #113, opened 2020-06-10
Fix doc hyperlinks to follow external site changes.

mtbc:
* LOCI's WiscScan software
* GE Healthcare now largely Cytiva.


sbesson:
Thanks now released as https://docs.openmicroscopy.org/ome-model/6.1.1/ome-tiff/index.html

PR #120, opened 2020-08-11
ome_model: channel fixes

sbesson:
This PR inclues a set of improvements for the `Image/Pixels/Channel` created by the `ome_model` library driven by in-progress IDR work on HCS RGB companion files /cc @dominikl 

## Changes

- the [Channel.Name](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Channel_Name) and [Channel.Color](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Channel_Color) attributes are optional as per the OME schema. In the case of many RGB images typically these values are irrelevant. 7837240  changes the `Channel` constructor and `Image.add_channel` method to turn these metadata fields into optional key/value pairs rather than mandatory arguments and handling missing arguments/`None`

-  1ae8b58  refines the current `Image.validate()` logic to account for the optional nature of the `Channel` element and the `SamplesPerPixel` value by using the two following checks:
  1. the number of `Channel` elements is less or equal to `SizeC`
  2. the sum of the `SamplesPerPixels` for each `Channel` is less of equal to `SizeC`

## Testing

Without this PR the following examples should fail during the channel validation step. With this PR, it should create a valid OME-XML:

```
from ome_model.experimental import Image, create_companion
image = Image("test", 512, 512, 1, 3, 1)
image.add_tiff("test.tiff")
create_companion(images=[image])
```

```
from ome_model.experimental import Image, create_companion
image = Image("test", 512, 512, 1, 3, 1)
image.add_channel(samplesPerPixel=3)
image.add_tiff("test.tiff")
create_companion(images=[image])
```

In addition to manual testing, commits f22547e and following add Python unit tests covering the constructors of the `Image` and `Channel` classes as well as a `tox` infrastructure used byTravis to run these Python tests in a separate job.

Target: `6.2.0`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 ---
+os: linux
+dist: bionic
 language: java
 
-sudo: false
-
-# http://docs.travis-ci.com/user/caching/#Arbitrary-directories
 cache:
   directories:
     - $HOME/.m2
@@ -13,25 +12,27 @@ addons:
     packages:
       - python3-pip
 
-jdk:
-  - openjdk11
-  - openjdk8
+before_install: pip3 install --user -r requirements.txt
 
-env:
-  - PYTHON=python3
-
-matrix:
+jobs:
   fast_finish: true
-
-before_install:
-  - python3 --version
-  - pip3 install --user -r requirements.txt
-
-install:
-  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dpython="$PYTHON"
-
-script:
-  - mvn test -B -Dpython="$PYTHON"
+  include:
+    - name: jdk8
+      language: java
+      jdk: openjdk8
+      install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dpython=python3
+      script: mvn test -B -Dpython=python3
+    - name: JDK11
+      language: java
+      jdk: openjdk11
+      install: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dpython=python3
+      script: mvn test -B -Dpython=python3
+    - name: Python36
+      python: "3.6"
+      install: pip3 install --upgrade pip wheel pytest tox virtualenv
+      script:
+        - ulimit -n 4096
+        - tox -e py36
 
 deploy:
   provider: pypi
@@ -40,4 +41,4 @@ deploy:
   distributions: sdist
   on:
     tags: true
-    jdk: openjdk8
+    python: "3.6"
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include LICENSE.md
+include pom.xml
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -47,7 +47,7 @@ class Channel(object):
         if name:
             self.data["Name"] = name
         if color:
-            self.data["Color"] = str(color),
+            self.data["Color"] = str(color)
         Channel.ID += 1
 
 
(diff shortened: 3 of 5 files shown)


dominikl:
Looks good, but with your second testing example I'll get:
```
AssertionError: {'Image': {'ID': 'Image:0', 'Name': 'test'}, 'Pixels': {'ID': 'Pixels:0:0', 'DimensionOrder': 'XYZTC', 'Type': 'uint16', 'SizeX': '512', 'SizeY': '512', 'SizeZ': '1', 'SizeT': '1', 'SizeC': '3'}, 'Channels': [<ome_model.experimental.Channel object at 0x7fe4463cd400>], 'TIFFs': [<ome_model.experimental.TiffData object at 0x7fe44559ab38>]}
```
And I can't find where/if the tests are running in the travis output, maybe I'm just missing what to look for?

dominikl:
I think it was correct, because `Image("test", 512, 512, 1, 1, 3)` now gets interpreted as `t=3`, doesn't it?

sbesson:
Sorry I meant the constructor imposes the order of the dimension size arguments (not the dimension order)

```
    def __init__(self,
                 name,
                 sizeX, sizeY, sizeZ, sizeC, sizeT,
                 tiffs=[],
                 order="XYZTC",
                 type="uint16",
                 ):
```

so `Image("test", 512, 512, 1, 1, 3)` creates an image with SizeZ=1, SizeC=1 and SizeT=3 - which is incompatible with a channel with 3 samples/pixels. 
but `Image("test", 512, 512, 1, 3, 1)` creates an image with SizeZ=1, SizeC=3 and SizeT=1 which can be a RGB image

sbesson:
Yes, tou are correct. The problem was that my channel samples comparison was too stringent (`<` while it should have been `<=`) and my test was not checking the case of 1 RGB channel only.
I have adjusted the test accordingly in https://github.com/ome/ome-model/pull/120/commits/2905ea9a01fed4f3dd47207822b183e304c27fdb so that it failed as expected by your testing and fixed the validation in https://github.com/ome/ome-model/pull/120/commits/49d212302bd30541e55ef64409b14377b069cd3a.

joshmoore:
Happy with a general code review. Also happy to open issue for "add ome_model version to output". I do wonder (similar to my recent ponderings around ome-zarr-py) if we don't want to hook this repository up to a MxN matrix of file producers/consumers, something like:

```
| V producer \ consumer > 	| bio-formats                     	| omero          	| ome-zarr-py                    	| omero-ms-zarr 	|
|-------------------------	|---------------------------------	|----------------	|--------------------------------	|---------------	|
| ome-model-py (TIFF)     	| Easy                            	| Easy           	| with bf2raw (needs ZarrWriter) 	| N/A (yet)     	|
| bio-formats (TIFF)      	| Easy                            	| Easy           	| "                              	| "             	|
| omero (TIFF Export)     	| Easy                            	| Easy           	|                                	| "             	|
| omero-cli-zarr (ZARR)   	| with raw2ometiff and ZarrReader 	| as bio-formats 	| Easy                           	| "             	|
| omero-ms-zarr (ZARR)    	| "                               	| "              	| Easy                           	| "             	|
|                         	|                                 	|                	|                                	|               	|
```

sbesson:
Thanks. Merging and I'll open 2 follow-up PRs: one for plane metadata and one for the version output.

Regarding compatibility, the closest official representation is the paragraph about tooling in https://docs.openmicroscopy.org/ome-model/6.1.2/. I could certainly see this turning into a producer/consumer or maybe read/write support matrix that we would keep updating as the community tooling grows.

I would assume we would need to discuss representation as I could imaging one table per container format ie. TIFF/ZARR.

PR #121, opened 2020-08-19
Ensure stdout works in Jupyter Notebooks

manics:
`sys.stdout` in Jupyter notebooks may not have a `buffer` attribute.

To test: run `create_companion(images=[companion])` (i.e. use the default `out` parameter) in a Jupyter Notebook.

See e.g. https://github.com/Yelp/mrjob/pull/1443

### DIFF
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -6,11 +6,6 @@ import sys
 import uuid
 import xml.etree.ElementTree as ET
 
-if sys.version_info[0] > 2:
-    PYTHON = 3
-else:
-    PYTHON = 2
-
 OME_ATTRIBUTES = {
     'Creator': "ome_model/experimental.py",
     'UUID': "urn:uuid:%s" % uuid.uuid4(),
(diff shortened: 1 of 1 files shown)


ome_model/experimental.py
```diff
@@ -6,11 +6,6 @@
 import uuid
 import xml.etree.ElementTree as ET
 
-if sys.version_info[0] > 2:
```
sbesson, line 9: :+1: especially as `setup.py` forces Python3+

ome_model/experimental.py
```diff
@@ -230,12 +225,12 @@ def create_companion(plates=[], images=[], out=None):
...
-    if PYTHON >= 3:
-        kwargs["xml_declaration"] = True
-        if not out:
+    kwargs["xml_declaration"] = True
+    if not out:
+        try:
             out = sys.stdout.buffer
-    elif not out:
-        out = sys.stdout
+        except AttributeError:
```
sbesson, line 232: Happy with either the `try/catch` approach or the `getattr` one-liner.

Is it possible to add a comment similar to https://github.com/Yelp/mrjob/pull/1443 explaining the special handling? Is there a reference page explaining why the stdout is different in the Jupyter notebook environment?

sbesson (changes requested):
Changes makes sense. Just one explanatory line about the context of the change would be useful. Otherwise LGTM.

Outside the scope of this PR, do you know of some mechanism allowing to test minimal Jupyter notebooks in a `tox` environment?

sbesson (approved):

ome_model/experimental.py
manics: Updated. I'm not aware of an explanation, my best guess is it was originally written to match python2 `sys.stdout` and no one though of adding `.buffer` to it.

manics:
For testing you can use nbval, for example
https://github.com/IDR/idr-notebooks/blob/master/docker/test_notebooks.sh
though note this script isn't yet automatically run in travis on that repo.

PR #122, opened 2020-08-20
ome-model: add support for Plane population

sbesson:
This API addition had  driven in the context of `idr0065` - see https://github.com/IDR/idr0065-camsund-crispri/blob/master/scripts/generate_companions.py#L146 to populate the `Plane` metadata using the submitted timestamp/exposure information - see https://idr.openmicroscopy.org/webclient/img_detail/9022301 for instance. As I was trying to re-use the API for `idr0092`, I realized I never backported the change.

In terms of style, the new `Plane` object explicitly defines the `z/c/t` arguments as well as an `options` dictionary which takes all the optional `Plane` attributes.   A minimal example of consumption of this API is illustrated below:

```
from ome_model.experimental import Image, create_companion

i = Image("test", 256, 512, 1, 2, 2)
i.add_plane(c=0, z=0, t=0, options={'DeltaT': '0.0'})
i.add_plane(c=0, z=0, t=1, options={'DeltaT': '1.0'})
i.add_plane(c=1, z=0, t=0, options={'DeltaT': '0.1'})
i.add_plane(c=1, z=0, t=1, options={'DeltaT': '1.1'})
create_companion(images=[i], out='planes.companion.ome')
```

The alternative is to define all these attributes explicitly in the constructor and the `Image.add_plane()` API. Feedback welcome.

### DIFF
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -48,14 +48,21 @@ class Channel(object):
 
 class Plane(object):
 
+    ALLOWED_KEYS = (
+        'DeltaT', 'DeltaTUnit', 'ExposureTime', 'ExposureTimeUnit',
+        'PositionX', 'PositionXUnit', 'PositionY', 'PositionYUnit',
+        'PositionZ', 'PositionZUnit')
+
     def __init__(self, TheC=0, TheZ=0, TheT=0, options={}):
         self.data = {
             'TheC': str(TheC),
             'TheZ': str(TheZ),
             'TheT': str(TheT),
         }
-        if options:
-            for key, value in options.items():
+        if not options:
+            return
+        for key, value in options.items():
+            if key in self.ALLOWED_KEYS:
                 self.data[key] = value
 
 
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -46,6 +46,19 @@ class Channel(object):
         Channel.ID += 1
 
 
+class Plane(object):
+
+    def __init__(self, TheC=0, TheZ=0, TheT=0, options={}):
+        self.data = {
+            'TheC': str(TheC),
+            'TheZ': str(TheZ),
+            'TheT': str(TheT),
+        }
+        if options:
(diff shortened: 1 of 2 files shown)


ome_model/experimental.py
```diff
@@ -46,6 +46,26 @@ def __init__(self,
...
+        'PositionX', 'PositionXUnit', 'PositionY', 'PositionYUnit',
+        'PositionZ', 'PositionZUnit')
+
+    def __init__(self, TheC=0, TheZ=0, TheT=0, options={}):
+        self.data = {
+            'TheC': str(TheC),
+            'TheZ': str(TheZ),
+            'TheT': str(TheT),
+        }
+        if not options:
```
joshmoore, line 62: nitpick: I'd tend to do `if options: ...` since a return in a constructor can be a surprise for someone who adds later logic.

ome_model/experimental.py
```diff
@@ -125,6 +146,13 @@ def add_tiff(self, filename, c=0, t=0, z=0, ifd=None, planeCount=None):
             planeCount=planeCount,
             uuid=UUID(filename)))
 
+    def add_plane(self, c=0, t=0, z=0, options={}):
+        assert c < int(self.data['Pixels']['SizeC'])
```
joshmoore, line 150: and `>= 0`?

joshmoore (commented):
Code looks generally good :+1: Few minor comments, both ignorable.

After https://github.com/ome/ome-model/pull/120#issuecomment-677514099, now I'm wondering if there couldn't be an integration test of sorts, perhaps using `tifffile` directly.

ome_model/experimental.py
sbesson: Fixed in a2819f8

ome_model/experimental.py
sbesson: Fixed in d2a7b15

joshmoore:
:+1: LGTM

PR #123, opened 2020-08-20
ome-model: allow to introspect version

sbesson:
Discussed with @manics and @joshmoore, this PR adds some logic to `setup.py` to write the current module version under `ome_model/version.py` using the `__version___` attribute - see https://www.python.org/dev/peps/pep-0396/. If `__init__.py` or another variable name is preferable, I have no strong opinion but I was not able to find a PEP authoritatively defining the recommended behavior.

The module is also updated to store `ome_model x.y.z[.dev0]` in the `OME.Creator` attribute similarly to the implementation in Bio-Formats - see https://github.com/ome/bioformats/pull/2207.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ target
 build
 dist
 ome_model.egg-info
-ome_model/version.py
+ome_model/__init__.py
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target
 build
 dist
 ome_model.egg-info
+ome_model/version.py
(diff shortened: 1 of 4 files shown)


joshmoore:
> to preserve backwards compatibility

Certainly for omero-web and omero-py these are part of the public API. I could see having the value that is imported into `<module>/__init__.py` be the thing we want to standardize on, but that doesn't solve the question of where to generate the value to.

sbesson:
Giving until tomorrow EOB for another round of discussions or alternate proposals and then I'll move forward with the release of `ome-model 6.2.0`.

PR #124, opened 2020-09-16
Convert Travis CI deployment into a build stage

sbesson:
The switch to build stages in #120 has broken the logic of the deployment stage so that `ome-model 6.2.0` was not automatically pushed to PyPI. This PR fixes the issue by moving the deployment to a dedicated build stage only executed on tags - see https://docs.travis-ci.com/user/build-stages/#build-stages-and-deployments

This re-uses most of the logic already implemented in https://github.com/ome/ome-zarr-py/blob/master/.travis.yml. https://travis-ci.org/github/sbesson/ome-model/jobs/727818756 also tests the conditional deployment logic (which failed due to the absence of credentials as expected)

Proposed tag `6.2.1`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,12 @@ jobs:
       script:
         - ulimit -n 4096
         - tox -e py36
-
-deploy:
-  provider: pypi
-  user: $PYPI_USER
-  password: $PYPI_PASSWORD
-  distributions: sdist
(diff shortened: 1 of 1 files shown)


joshmoore (approved):
Screen Shot 2020-09-17 at 11 41 35
:+1:

PR #125, opened 2020-10-19
OMEModelImpl: Do not permit the addition of null references

rleigh-codelibre:
This is a small fix I was going to make for OME Files (see [MR !116](https://gitlab.com/codelibre/ome/ome-model/-/merge_requests/116/diffs?w=1)).  It prevents the addition of null references to the model which would cause problems later on when traversing the model data after the initial import.  Should not be seen in normal use, but it protects against unintentional incorrect use.  Returns a failure status as for other failure scenarios, so the caller will be notified the addition failed should they care to check.


dgault:
Makes sense to return a failed status rather than allow the addition.

PR #86, opened 2018-09-27
Add gradle build

rleigh-codelibre:
--exclude

This PR adds a gradle build system to replace the existing maven build.  Compatible with #65. The maven build is left in place for now to avoid conflicts. CI tests are now using gradle.  gradle and openjdk7 aren't compatible on travis (see https://github.com/travis-ci/travis-ci/issues/8503), so test with 8 and 10, dropping 7.  The source version could be upgraded to 1.8, but this has been left for a future PR.

This should be a drop-in replacement for all of the maven functionality in current use, and it's vastly shorter and simpler than the maven logic.  This is the most complex bioformats-related package, with multiple components, source code generation and documentation generation, so this conversion is as complex as it will get for all of the decoupled components.  It includes all the needed logic for building, testing and making signed releases.

Testing:

Run `./gradlew build|assemble|test|publishToMavenLocal`

To test packaging changes:

```
mvn
./gradlew assemble
pkgdiff ome-xml/target/ome-xml-5.6.4-SNAPSHOT.jar ome-xml/build/libs/ome-xml-5.6.4-SNAPSHOT.jar
(repeat for other jars, and look at the HTML pages for changes; should be cosmetic only--pom.xml/pom.properties not in jar [maven-specific] and some minor MANIFEST.MF differences)
```

Workflow changes:

- Edit `VERSION` to set the version number.  Used by both the gradle and cmake builds.
- Create `~/.gradle/gradle.properties` (equivalent of `~/.m2/settings.xml`) with this content:

```
signing.keyId=<keyid>
ossrhUsername=<username>
ossrhPassword=<password>
```

- Run `./gradlew publish` to upload to maven central.  Snapshots to snapshots, releases to staging handled automatically.  Or run `./gradlew publishSnapshot` to upload snapshots only, skipping releases, which is useful for CI builds where signing is not possible.

PR #129, opened 2020-11-24
Add gradle build

rleigh-codelibre:
Follow up for #86.  This PR no longer has any C++-related conflicts, and is updated to the latest stable gradle release, 6.7.1 and the latest ome-model master version.  This should keep you up-to-date with the possibilities even if you don't want to merge it at this point in time.

--exclude

This PR adds a gradle build system to replace the existing maven build.  CI tests are now using gradle, testing with JDK 8 and 11.  The source and target versions are 1.8.

This should be a drop-in replacement for all of the maven functionality in current use, and it's vastly shorter and simpler than the maven logic.  This is the most complex bioformats-related package, with multiple components, source code generation and documentation generation, so this conversion is as complex as it will get for all of the decoupled components.  It includes all the needed logic for building, testing and making signed releases.

Testing:

Run `./gradlew build|assemble|test|publishToMavenLocal`

To test packaging changes:

```
mvn
./gradlew assemble
pkgdiff ome-xml/target/ome-xml-6.2.2-SNAPSHOT.jar ome-xml/build/libs/ome-xml-6.2.2-SNAPSHOT.jar
(repeat for other jars, and look at the HTML pages for changes; should be cosmetic only--pom.xml/pom.properties not in jar [maven-specific] and some minor MANIFEST.MF differences)
```

Workflow changes:

- Edit *VERSION* to set the version number.  Was originally done to accommodate Java and C++/CMake, now used to accommodate Java and Python setup.py; can be easily moved into the gradle properties if desired.
- Create *~/.gradle/gradle.properties* (equivalent of *~/.m2/settings.xml*) with this content:

```
signing.gnupg.keyName=<keyid>
ossrhUsername=<username>
ossrhPassword=<password>
```

You might also want:

```
python=python3
```

(can be put into *gradle.properties* in ome-model to customise the single project for your system).

- Run `./gradlew publish` to upload to maven central.  Snapshots to snapshots, releases to staging handled automatically.  Or run `./gradlew publishSnapshot` to upload snapshots only, skipping releases, which is useful for CI builds where signing is not possible.


rleigh-codelibre:

It's possible that the GitHub Actions support can be further improved to cache the Gradle download and other dependencies if they aren't already being taken care of by caching *~/.m2*.

rleigh-codelibre:
Note that this will conflict with #127 and will need updating to also include Python 3.9 support.

sbesson:
Briefly discussed at the Formats meeting. Although there is no ruling out that we might reconsider Maven as the build system at some point in the future, there is no immediate driver on the OME side. Rather than keeping this open perpetually, closing it and we can always resurrect the initial proposal as needed if we re-evaluate the build system.

Issue #130, opened 2020-11-26
MetadataConverter: WellAnnotationRef conversion defect


rleigh-codelibre:
The Java MetadataConverter does not appear to be handing `WellAnnotationRef` correctly.  `wellAnnotationRefCount` is never assigned any value other than 0.  Fix appears straightforward.

Impact: All `WellAnnotationRef`s are silently dropped during this conversion.

The OME-Files C++ converter does not contain this defect.

rleigh-codelibre:
I'm testing a fix for this, and I'll open a PR once it's done.

PR #131, opened 2020-11-27
MetadataConverter: Fix dropping of WellAnnotationRefs during conversion

rleigh-codelibre:
Correctly initialise wellAnnotationRefCount.

This defect was found during static analysis of the source tree.  I will have several additional fixes to come, plus a number of performance and cosmetic improvements also picked up by the static analyser.

Testing: Try converting an OME-TIFF with `WellAnnotationRef`s.  You should see that they are stripped out during conversion.  Repeat with this PR.  You should find they are now correctly preserved.

Closes: #130

### DIFF
--- a/ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
+++ b/ome-xml/src/main/java/ome/xml/meta/MetadataConverter.java
@@ -2301,7 +2301,7 @@ public final class MetadataConverter {
 
         int wellAnnotationRefCount = 0;
         try {
-          src.getWellAnnotationRefCount(i, q);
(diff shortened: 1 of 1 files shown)


rleigh-codelibre:

sbesson (approved):
Straightforward fix tested with this variation of the sample screen OME-XML with two wells annotated with a `TagAnnotation`
[well_annotation.ome.xml.zip](https://github.com/ome/ome-model/files/6689710/well_annotation.ome.xml.zip).

Without this PR, the annotation reference is lost during the metadata handling either using `showinf` or `bfconvert`. With this PR included the annotation reference is preserved as expected.

Issue #132, opened 2020-11-27
EnumHandlers: Failure to check types correctly


rleigh-codelibre:
For each of the generated EnumHandler classes, the types are evaluated in the wrong order.  None of the `Positive(Integer|Long|Float)` types will be used because they are extending the `NonNegative(Integer|Long|Float)` types and these are checked first, and will match the `Positive` cases too.

Should be fixable by changing the ordering.

PR #133, opened 2020-11-27
OMEXMLModelEnumHandler template: Correct type evaluation order

rleigh-codelibre:
Closes: #132 

Testing: You will need to check which type each primitive type extends and confirm that the ordering is from most-derived to least-derived so that all types are correctly processed.  You could alternatively run a static analyser on the generated sources to confirm that none of the conditional blocks are unreachable (which is how the defect was originally found).

Generated source changes: Download the diff from [here](https://gitlab.com/rleigh/ome-model/-/jobs/875951611) or generate the diff yourself.


rleigh-codelibre:

sbesson (approved):

PR #134, opened 2020-11-27
OMEXMLModelEnum template: Correct exception handling

rleigh-codelibre:
Static analysis highlighted a bad use of `return` within a `finally` block.  This was indeed unnecessary and replaceable with an ignored `catch` block.

Testing: Nothing specific required; the effective code flow is unchanged with or without exceptions, but we no longer return from within a `finally` block.

Generated source changes: Download the diff from [here](https://gitlab.com/rleigh/ome-model/-/jobs/875962162) or generate the diff yourself.


rleigh-codelibre:

sbesson (approved):

PR #136, opened 2020-11-29
Static analysis code remediation: ome-xml and specification Java sources

rleigh-codelibre:
Following static analysis of the ome-model source tree, the following remediation of the ome-xml and specification Java components has been performed:

Both ome-xml and specification:

* Addition of missing `final` qualifiers
* Removal of inferrable types when using generics (replaced with `<>`)
* Remove unnecessary boxing and unboxing

ome-xml:

* Simplify complex conditional
* Make ignored exceptions explicit
* Replace complex sets of conditionals with switch
* Use `Map.computeIfAbsent()` to simplify list creation
* Remove `public` modifier from interface (it's public by default)
* Replace use of `StringBuilder` for performance
* Return values directly rather than using an unneeded temporary
* Add suppression for unused return values [within the ome-model source tree]
* Add suppression for same return values for functions returning constants

specification:

* Remove redundant null initialisers (when the local will always be initialised immediately after creation and before first use)
* Remove unneeded temporary values
* Use try-with-resources rather than performing manual clean-up
* Correction of Javadoc markup
* Removal of unused exception types
* Use of new-style for loops
* Use of StandardCharsets to replace string literal

No generated source changes.

Total static analysis warnings fixed: 702 (3261 before, 2559 after)

Testing:

The refactoring should not result in any behavioural changes. All unit tests and integration tests should continue to pass. There should be no compatibility changes for downstream consumers.

If you would prefer to review these changes in a more fine-grained way, it can be split out into multiple PRs without any trouble. Few of the commits depend upon each other, so can be freely cherry-picked.

### DIFF
--- a/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
+++ b/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
@@ -990,7 +990,7 @@ public class XMLMockObjects
     plate.setWellOriginY(new Length(1.0, UNITS.MICROMETER));
     plate.setStatus("Plate status");
     PlateAcquisition pa = null;
-    List<PlateAcquisition> pas = new ArrayList<PlateAcquisition>();
+    List<PlateAcquisition> pas = new ArrayList<>();
     int v;
     if (numberOfPlateAcquisition > 0) {
       for (int i = 0; i < numberOfPlateAcquisition; i++) {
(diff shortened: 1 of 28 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [BIOFORMATS-push#659](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/659/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-push/659/consoleText) for more details.
Possible conflicts:
  - PR #125 rleigh-codelibre 'OMEModelImpl: Do not permit the addition of null references'
    - ome-xml/src/main/java/ome/xml/model/OMEModelImpl.java

--conflicts

rleigh-codelibre:
Update pushed which fixes up some cosmetic formatting in `MetadataConverter`.

snoopy is correct that #125 does conflict.  As and when that fix is merged, I'll update this PR to remove the conflict.

rleigh-codelibre:

melissalinkert (approved):
I don't see anything obviously problematic in the diff, and this has been successfully included in merge builds for a long time now.

PR #137, opened 2020-11-30
change link

jburel:
point to a validator now that the previous one is gone
I picked a W3C XML Schema (XSD) Validation online 
If someone has a better on in mind, please suggest one

### DIFF
--- a/docs/sphinx/ome-tiff/tools.rst
+++ b/docs/sphinx/ome-tiff/tools.rst
@@ -155,10 +155,8 @@ are only 2 mistakes in the file.
    -  correct TiffData block usage
    -  frame counts in TIFF files
 
-Alternatively, you can use the more general online `W3C XML
-validator <http://www.utilities-online.info/xsdvalidation>`_ to validate your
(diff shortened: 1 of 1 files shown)


docs/sphinx/ome-tiff/tools.rst
```diff
@@ -156,7 +156,7 @@ are only 2 mistakes in the file.
    -  frame counts in TIFF files
 
 Alternatively, you can use the more general online `W3C XML
-validator <http://www.w3.org/2001/03/webdata/xsv>`_ to validate your
+validator <http://www.utilities-online.info/xsdvalidation>`_ to validate your
```
sbesson, line 159: Definitely a tricky one. The link you mentioned is certainly an alternative but raises the question of the maintenance. 

The message from https://www.w3.org/XML/ is very much in the spirit of "many tools exist, search and pick one".  My only other suggestion would be to rewrite the last two paragraphs to mention OME-XML can be validated used any valid XML tool.

jburel: I think  we should not suggest one and indicate that several tools exist

PR #138, opened 2020-12-07
xsd-fu: Remove Genshi

rleigh-codelibre:
This PR removes the embedded copy of Genshi and adds it to *requirements.txt*.  I'll do some further testing on a variety of platforms to assess the impact.


rleigh-codelibre:
Please see [MR !185](https://gitlab.com/codelibre/ome/ome-model/-/merge_requests/185) and the corresponding [CI pipeline](https://gitlab.com/rleigh/ome-model/-/pipelines/226404263).

I did need to update some of the CI container images and a few of the build nodes to make sure Genshi was properly provisioned, but following that it has been tested on FreeBSD 12.2, MacOS 11 and 10.5, CentOS 8, Ubuntu 20.04 and Windows with VS2017 and VS2019 (all C++) and Ubuntu 20.04 with Java 11 and Windows with Java 11.  No problems noted.

I think it's safe to merge.  The main risk is future breakage if an incompatibility arises.  The other is on systems which package Python modules but do not include Genshi.  This includes CentOS 8, FreeBSD and MacOS.  In all cases, it is installable via pip.  But the assumption being made is that the version provided by pip is functional, and while it is true today, it may not be the case following future Python changes.  I don't think that should block merging, however.  The worst case is that we will need to revert back to the embedded copy, and that's pretty trivial should the need arise.

sbesson (approved):
Thanks @rleigh-codelibre. Confirmed the removal of the embedded Genshi copy does not destabilize any of the integration builds or the repository tests. I think the usage of the upstream is fine with the caveats mentioned above.

Merging this and I will cut a new patch release of `ome-model`.

PR #143, opened 2021-05-20
Add physical sizes and units to Image class

ehrenfeu:
As discussed in the [OMERO custom plate import][1] thread on the Image.sc forum.

[1]: https://forum.image.sc/t/omero-custom-plate-import/52381/14

### DIFF
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -105,6 +105,10 @@ class Image(object):
                  tiffs=[],
                  order="XYZTC",
                  type="uint16",
+                 physSizeX="1.0", physSizeY="1.0", physSizeZ="1.0",
+                 unitX="px",
+                 unitY="px",
+                 unitZ="px",
                  ):
         self.data = {
             'Image': {'ID': 'Image:%s' % self.ID, 'Name': name},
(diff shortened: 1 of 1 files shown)


ome_model/experimental.py
```diff
@@ -105,6 +105,10 @@ def __init__(self,
                  tiffs=[],
                  order="XYZTC",
                  type="uint16",
+                 physSizeX="1.0", physSizeY="1.0", physSizeZ="1.0",
```
sbesson, line 108: Primary question is about the extensibility and the consistency of the API as we want to make more optional properties available to `Image` but also other objects. The two options I have been debating previously are:

- support for each new attribute as a key/value pair as done here
- passing an optional dictionary e.g. https://github.com/ome/ome-model/blob/38624e963f4dbde67992416d000d9eef26828f3b/test/unit/test_image.py#L254-L267. 

My preference goes for the second form as it enforces the name of the attributes that can be passed to the optional dictionary to match the XSD schema.

sbesson (commented):
Overall, strong endorsement to allow physical sizes to be passed to the constructor in this utility API.

My main consideration is about the usage of explicit key/value pairs vs a single extra dictionary containing the extra attributes. Once settled, the only other thing worth adding would be some unit tests to ensure these attributes are correctly written to disk.

Issue #142, opened 2021-05-17
OriginalMetadata from 2003-FC lost in transform process


dgault:
Issue is a follow up to image.sc thread https://forum.image.sc/t/bio-formats-returns-all-metadata-except-ome-meta-data/52295/2

Using the sample file provided in the thread and correcting validation errors shows that the original metadata (stored in 2003-FC  spec) gets lost when going through the transform update process. The particular transform the issue occurs with is 
`2008-09-to-2009-09.xsl`

The data that gets lost is stored using the `OriginalMetadata` tag such as below (adding the CA namespace didn't resolve the issue):
```
<CustomAttributes>
<OriginalMetadata ID="OriginalMetadata:3" Name="Acquisition Parameters/Channel 2" Value="Fluorescence"/>
</CustomAttributes>
```

In the transform this should be handled by the following, transforming the metadata to an XMLAnnotation:
```
  <xsl:template match="CA:*">
    <xsl:element name="{name()}" namespace="{$newCANS}">
      <xsl:apply-templates select="@*|node()"/>
    </xsl:element>
  </xsl:template>

  <!-- Transform the CustomAttributes into XMLAnnotation -->
  <xsl:template match="CA:CustomAttributes">
    <xsl:if test="count(@*|node()) &gt; 0">
      <xsl:element name="StructuredAnnotations" namespace="{$newSANS}">
        <xsl:element name="XMLAnnotation" namespace="{$newSANS}">
          <xsl:attribute name="ID">Annotation:1</xsl:attribute>
          <xsl:element name="Value" namespace="{$newSANS}">
            <xsl:apply-templates select="@*|node()"/>
          </xsl:element>
        </xsl:element>
      </xsl:element>
    </xsl:if>
  </xsl:template>
```

dgault:
As a follow up to this a PR has been opened which aims to resolve this bug (https://github.com/ome/ome-model/pull/144) along with some new unit tests showing the original failure (https://github.com/ome/bioformats/pull/3695).

PR #144, opened 2021-05-21
2008-09 to 2009-09: Move CustomAttributes transforms to top level

dgault:
This is a follow up to https://github.com/ome/ome-model/issues/142 and should see the unit tests in https://github.com/ome/bioformats/pull/3695 turn green.

The issue appears to have been due to `CustomAttributes` being able to be located in 4 different locations:
- top level under OME
- under images
- under images/feature
- under dataset

The non top level `CustomAttributes` were being transformed into `XMLAnnotations` but embedded in the original element, be it Image or Dataset, this is an invalid location for it and thus it was not being recognised when converted back to an OME-Model object. This PR aims to move all `CustomAnnotations` to the top OME level before transforming them. If you have multiple `CustomAttributes` at different levels then you will get multiple `XMLAnnotations` inside a single `StructuredAnnotations`.

Fixes https://github.com/ome/ome-model/issues/142

### DIFF
--- a/specification/src/main/resources/transforms/2008-09-to-2009-09.xsl
+++ b/specification/src/main/resources/transforms/2008-09-to-2009-09.xsl
@@ -154,6 +154,22 @@
         </xsl:element>
       </xsl:for-each>
 
+      <!-- Transform the CustomAttributes into XMLAnnotation -->
+      <xsl:if test="(count(//*[local-name() = 'CustomAttributes'])) &gt; 0">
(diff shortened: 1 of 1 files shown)


sbesson:
@dgault looks like this PR has been opened from a branch on the `ome/ome-model` repository itself. Unfortunately, while Travis has been disabled for Pull Requests, there was a remaining webhook that created a failing build on the push event which will I cannot easily remove.

Could you either force push or recreate a PR from your fork?

PR #147, opened 2021-06-16
Publish doc on rtd

jburel:
Build the doc on readthedocs

This is currently available at https://readthedocs.org/projects/jburel-ome-model/
The link in the README will need to be updated when we have the final url

https://docs.readthedocs.io/en/stable/versions.html#states for version
I have also not activated the webhooks.

cc @sbesson


sbesson (commented):
Overall, I do like the rendered output especially if we are moving all the documentation over to the unified RTD theme. I reviewed a few pages and the content is identical to the one currently published under docs.openmicroscopy.org/ome-model/

A few high-level questions to discuss with @ome/formats in order to move forward:

- the left-side menu lists all the individual pages. It would be useful to have the structure defined in the index (OME-TIFF, OME-XML, ...)
-the version/release in the documentation title is a  confusing. With the proposed `latest/stable` theme, would the published documentation not always be a SNAPSHOT/dev0 version? 
- one of the major change is the removal of `docs/sphinx` as a Maven module and the impact on the build system. The immediate effect is that this breaks some assumptions of the CI jobs - see https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-linkcheck/501/ and https://merge-ci.openmicroscopy.org/jenkins/job/OME-MODEL-linkcheck/351/ which will need to be reviewed. A secondary impact is that this documentation would not longer be built by https://github.com/ome/bio-formats-build which relies on the Maven coupling.

Pushing the last point further, the spirit of the change proposed in this PR further reduces increases the divergence between `docs/sphinx` and the other components of this repositories. For me, this raises the question of whether we should  extract the documentation to its own repository and let it be maintained and deployed independently.

jburel:
for the version: this the current one in the file. When a tag is pushed the version X.X.X will be shown. 
I think we should consider hiding latest i.e. the dev work.
The CI infra will certainly need some TLC, I should have mentioned that in the description.

Having the doc independently will make sense also the layout used here ``docs/sphinx`` is not used in other repositories

sbesson:
Discussed today. The current consensus seems to lean toward extracting this documentation and maintaining it separately. @jburel will come back to this when possible. In the meantime, excluding it to reduce the build failures.

jburel:
The release of the documentation is still done via https://ci.openmicroscopy.org/job/OME-MODEL-release
We are shutting down ci.o.org.
We need to get back to this work

PR #149, opened 2021-06-22
ome_model: do not use mutable default arguments

sbesson:
Triggered by an IDR submission, the following code exposes the issue

```python
from ome_model.experimental import Plate, Image, create_companion

for i in range(5):
    plate = Plate(str(i), 1, 1)
    well = plate.add_well(0,0)
    image = Image(
      str(i), 10, 10, 1, 2, 1,
                order="XYZCT", type="uint16")
    image.add_tiff(str(i) + '.tif')
    well.add_wellsample(0, image)
    create_companion(plates=[plate], out=str(i) + '.companion.ome')
```

While expecting this will create 4 companion OME files with 1 plate/1 well/1 well sample/1 image each, the second companion contains 2 images, the third 3...

Tracing down the source of the issue comes down to the usage of empty lists as default arguments which is a classical Python gotcha as these are mutable. At the moment, a workaround would be to explicitly pass a new empty list.

```diff
-    create_companion(plates=[plate], out=str(i) + '.companion.ome')
+    create_companion(plates=[plate], images=[], out=str(i) + '.companion.ome')
```

87f4859 updates all locations using empty lists as default to using `None` instead and handles the check within the method

All the other commits add some unit test coverage of the plate generation scenarios. The tests should be failing without 87f4859 and passing with.

### DIFF
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -102,7 +102,7 @@ class Image(object):
     def __init__(self,
                  name,
                  sizeX, sizeY, sizeZ, sizeC, sizeT,
-                 tiffs=[],
+                 tiffs=None,
                  order="XYZTC",
                  type="uint16",
                  ):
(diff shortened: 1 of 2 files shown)


joshmoore (approved):
`()` would be another alternative, but yeah, a typical gotcha. Not sure why a linter didn't catch it.

PR #151, opened 2021-10-06
Add pixel sizes

dominikl:
Add the possibility of adding the pixel sizes to the companions. Also removes two unused Plate constructor variables.

### DIFF
--- a/ome_model/experimental.py
+++ b/ome_model/experimental.py
@@ -102,6 +102,7 @@ class Image(object):
     def __init__(self,
                  name,
                  sizeX, sizeY, sizeZ, sizeC, sizeT,
+                 pixX=None, pixY=None, pixZ=None,
                  tiffs=None,
                  order="XYZTC",
                  type="uint16",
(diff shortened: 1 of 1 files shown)


ome_model/experimental.py
```diff
@@ -167,7 +174,7 @@ class Plate(object):
 
     ID = 0
 
-    def __init__(self, name, rows, columns):
+    def __init__(self, name):
```
sbesson, line 177: That one is likely on me as this was introduced in https://github.com/ome/ome-model/pull/92 but never implemented. 

I suspect in the back of my mind, I was thinking of the `Plate.Rows` and `Plate.Columns` fields (see https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Plate) but never got to implement them.

Rather than deleting the arguments which is a breaking change (as you are updating the test), thoughts about making them optional e.g. `rows=None, columns=None`) and optionally filling the fields?

ehrenfeu:
Now if only I remembered where this will affect my code... 😳

ehrenfeu:
Thanks @sbesson - I found the original thread already, now I only need to find the time to dig into the stuff on our side...

sbesson:
@dominikl @ehrenfeu this is now available as https://pypi.org/project/ome-model/6.3.0/

PR #154, opened 2021-12-14
Migrate XSLT transformation page

sbesson:
Following the deployment work in https://github.com/ome/prod-playbooks/pull/329, https://www.openmicroscopy.org/Schemas/Transforms/ is now returning a 403. It was previously returning a listing of the directory content auto-generated by the web server but a proper landing page had never been put into place. The model documentation is the only place in the OME stack which was linking to this endpoint.

Separately, a long-standing issue has been that the XSLT transformations was currently maintained in the OMERO documentation (https://docs.openmicroscopy.org/omero/5.6.3/developers/Model/XsltTransformations.html) to document the upgrade/downgrade quality.

This PR proposes to address both issues by removing the URL link and adding a new transformations page to the Sphinx documentation migrating most of the content in OMERO. It also includes the information for the 2016-06 <-> 2015-01 transformations. The last table (transformation paths) is currently dropped.


jburel:
The content is there. 
The layout of the table is not great. It should be encapsulated in a scrollable pane if possible.
Screenshot 2022-04-20 at 09 51 18

sbesson:
For additional background, I started with https://docs.openmicroscopy.org/omero/5.6.3/developers/Model/XsltTransformations.html#quality-of-transformations where I realised that:
- the table is an exported image from a spreadsheet (i.e. hard to modify)
- the current version does not include the current version of the model (2016-06)
- the table in the OMERO documentation is already fairly small so adding another column/row will make it less usable

So far, I went for the simplest approach of turning this image into a reStructuredText table to help the content management/searching but I agree it suffers from a readability perpective

A few questions:
- can you think of another representation that would better suit this content?
- assuming a table remains as the best representation, would a horizontal scrollbar suffice?

Incidentally, I noticed that https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html#list-tables has some built in scrollbar elements so possibly this might be worth evaluating on top of #147 ?
If this was the preferred approach, I would also propose to invert the order of the rows/columns so that the most recent versions show up in the top-left corner.

joshmoore:
> can you think of another representation that would better suit this content?

In my opinion (or rather, in my use for presentations) the color was probably more important than the text if that helps, which is where this came from:
Screen Shot 2022-04-20 at 21 13 24

`Transform "drawing"`

jburel:
The color approach is definitely more suited.

joshmoore:
Semi-tongue in cheek, I assume we could export XLS to SVG. 🙂

sbesson:
> Semi-tongue in cheek, I assume we could export XLS to SVG. 🙂 

😄 I was expecting nothing less from you 

If you guys agree this is the best strategy in terms of display, happy to consolidate by
1- moving the sources as xls directly under this repo (maybe also CSV, but it will lack the color scheme), 
2- export as SVG for inclusion in the page

sbesson:
My brief attempts at doing XSLS -> SVG have not really resulted in the expected output, either because the text is converted to graphic elements or because the online converter I used generated some extra dependencies like fonts. 
@joshmoore @jburel would you be happy for this to be merged as it is as a step-wise improvements so that we have a page that we can reference from the OMERO documentation and elsewhere?

jburel (approved):
I think it is an improvement. We can improve and iterate.
I am for merging this first step of the work

PR #158, opened 2022-04-15
XMLMockObjects: make artificial SHA1 schema-compliant

melissalinkert:
See https://github.com/ome/bioformats/issues/3810

### DIFF
--- a/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
+++ b/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
@@ -1093,7 +1093,6 @@ public class XMLMockObjects
     plane.setTheZ(new NonNegativeInteger(z));
     plane.setTheC(new NonNegativeInteger(c));
     plane.setTheT(new NonNegativeInteger(z));
-    plane.setHashSHA1("1234567890ABCDEF1234");
     return plane;
   }
 
--- a/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
+++ b/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
@@ -1093,7 +1093,7 @@ public class XMLMockObjects
     plane.setTheZ(new NonNegativeInteger(z));
     plane.setTheC(new NonNegativeInteger(c));
     plane.setTheT(new NonNegativeInteger(z));
-    plane.setHashSHA1("1234567890ABCDEF1234567890ABCDEF12345678");
+    plane.setHashSHA1("1234567890ABCDEF1234");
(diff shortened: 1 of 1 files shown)


sbesson:
To my original surprise, this change failed the unit tests in https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-image/1151 with the following validation error

```
2022-04-18 00:17:05,448 [main] ERROR loci.common.xml.XMLTools - cvc-length-valid: Value '1234567890ABCDEF1234' with length = '10' is not facet-valid with respect to length '20' for type 'Hex40'.
2022-04-18 00:17:05,449 [main] ERROR loci.common.xml.XMLTools - cvc-type.3.1.3: The value '1234567890ABCDEF1234' of element 'HashSHA1' is not valid.
```

After some investigation, I suspect the explanation is that the `HashSHA1` element is of `Hex40` type defined as

https://github.com/ome/ome-model/blob/e6efb8802492880f8fde7be99c1c9b83be215e76/specification/src/main/resources/released-schema/2016-06/ome.xsd#L1467-L1476

So the 20 length restriction might not apply to the hexadecimal string but rather to its equivalent binary representation. This matches what Python `binascii.unhexlify` returns about the former value and the newer one


```
>>> from binascii import unhexlify
>>> len(unhexlify("1234567890ABCDEF1234567890ABCDEF12345678"))
20
>>> len(unhexlify("1234567890ABCDEF1234"))
10
>>> unhexlify("1234567890ABCDEF1234567890ABCDEF12345678")
b'\x124Vx\x90\xab\xcd\xef\x124Vx\x90\xab\xcd\xef\x124Vx'
```

melissalinkert:
Following today's formats team discussion, removing `HashSHA1` entirely from `XMLMockObjects` and re-including.

The longer-term solution is to update the schema to deprecate `HashSHA1`, and ensure that nothing else in our stack uses this feature. `git grep HashSHA1` on https://github.com/ome/bioformats/tree/3d73a85c5af636cf5af42e952adae4e22611585b shows nothing.

sbesson (approved):
The original issue associated with the reading the Plane.HashSHA1 has been fixed downstream - see e.g. https://github.com/tlambert03/ome-types/pull/146.

As discussed at the Formats meeting, a secondary problem is that the Plane.HashSHA1 carries the expectation to store a hash of the plane data. At the moment, this mock infrastructure will unconditionally store an arbitrary value which has the potential to create confusion for downstream consumers.

Given all the builds have been staying green with this PR included, I would agree to get it merged and included in the next ome-model release.

PR #159, opened 2022-05-15
Add Java 17 to the CI build matrix

sbesson:
Also drop Python 3.6 (EOL) and only test different Python versions once per platform

Tests should remain green

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -14,6 +14,15 @@ jobs:
         java: [1.8, 11, 17]
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ['3.7', '3.8', '3.9']
+        exclude:
+          - java: 11
+            python-version: '3.8'
+          - java: 11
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -11,9 +11,18 @@ jobs:
   build:
     strategy:
       matrix:
-        java: [1.8, 11]
+        java: [1.8, 11, 17]
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.6', '3.7', '3.8', '3.9']
+        python-version: ['3.7', '3.8', '3.9']
```
joshmoore, line 16: I assume it's not overly relevant for this repo, but soon I could imagine adding 3.10 as well.

sbesson: All fine with me,  shifting the supported Python version range would go in pair with the dropping of EOL Python 3.6. Feel free to open a PR to that effect

PR #165, opened 2022-09-28
BBBC017: new representative OME-TIFF samples

sbesson:
This PR updates the sample data page of the OME-TIFF specification to point at the new representative samples of the BBBC017 plate which were recently generated and published under https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/BBBC017/.

These samples are being tested by the daily CI builds and have been used for the functional testing of https://github.com/ome/bioformats/pull/3831#pullrequestreview-1100874964

A note is also added to explain why the former OME-TIFF should be considered as deprecated and its usage replaced.


melissalinkert (approved):

dgault (approved):
Updated text looks good to me and links are correct

PR #167, opened 2023-01-10
Add Media Cybernetics

pwalczysko:
Adding a new member of the ``supported`` list on https://docs.openmicroscopy.org/ome-model/latest/ome-tiff/index.html 

cc @dgault @jrswedlow

### DIFF
--- a/docs/sphinx/conf.py
+++ b/docs/sphinx/conf.py
@@ -361,4 +361,5 @@ linkcheck_ignore = [
     "http://www.perkinelmer.com/", # 500 Error
     "http://mayachitra.com/", # SSL certificate expired
     "https://www.cytivalifesciences.com/",
+    "https://www.mediacy.com/", # https://www.digicert.com/ does not trust the issuer, but link is valid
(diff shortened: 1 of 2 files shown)


pwalczysko:
Although the commit seems to be included in https://github.com/snoopycrimecop/ome-model/blob/merge_ci/docs/sphinx/conf.py but the build https://merge-ci.openmicroscopy.org/jenkins/job/OME-MODEL-linkcheck/943/console is still red ?

Any ideas please ? Maybe it will magically solve itself tomorrow ? 

cc @joshmoore

pwalczysko:
Brilliant, so if I run the https://merge-ci.openmicroscopy.org/jenkins/job/BIOFORMATS-image/ after the BF-push and BF-build jobs, then the test commit gets taken into account - got it !

Hence, removing the test commit and hoping that all will be green tomorrow.

dgault (approved):
Link looks correct to me, we also have it linkcheck ignore for the main bioformats repo

PR #171, opened 2023-04-06
Update specification.rst

Tom-TBT:
Hello,
I spotted what I think is an error in one of the examples of the doc, which confused me enough to make this very little PR.

Best,
Tom

### DIFF
--- a/docs/sphinx/ome-tiff/specification.rst
+++ b/docs/sphinx/ome-tiff/specification.rst
@@ -366,7 +366,7 @@ example the identifiers could be distributed as follows:
             </UUID>
           </TiffData>
     ...
-          <TiffData FirstC="0" FirstT="1" FirstZ="0" IFD="1" PlaneCount="1">
+          <TiffData FirstC="0" FirstT="1" FirstZ="0" IFD="0" PlaneCount="1">
             <UUID FileName="tubhiswt_C0_TP1.ome.tif">
               urn:uuid:743275b7-6726-46bd-b7bb-aca3085f429a
             </UUID>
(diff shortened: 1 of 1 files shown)


Tom-TBT:
For context, this is here: https://docs.openmicroscopy.org/ome-model/6.3.1/ome-tiff/specification.html#embedded-ome-xml-metadata

and the change corresponds to set the value under the red arrow to 0:
image

docs/sphinx/ome-tiff/specification.rst
```diff
@@ -366,7 +366,7 @@ example the identifiers could be distributed as follows:
             </UUID>
           </TiffData>
     ...
-          <TiffData FirstC="0" FirstT="1" FirstZ="1" IFD="1" PlaneCount="1">
+          <TiffData FirstC="0" FirstT="1" FirstZ="0" IFD="1" PlaneCount="1">
```
dgault, line 369: Thanks @Tom-TBT for spotting this, it does make more sense to have it showing the first IFD. I have added a suggestion to also update the IFD index as well as the Z index.

```suggestion
          <TiffData FirstC="0" FirstT="1" FirstZ="0" IFD="0" PlaneCount="1">
```

dgault (changes requested):

docs/sphinx/ome-tiff/specification.rst
Tom-TBT: Was it acceptable then to have FirstZ="1" with IFD="1"   ?

With this example, I was looking if TiffData was the same between `tubhiswt_C0_TP0.ome.tif` and `tubhiswt_C0_TP1.ome.tif`, but I didn't notice that the IFD was different. 

Sorry for taking on your time for such small details

dgault (approved):

PR #180, opened 2023-09-27
xsd-fu: templates for linkml generation

joshmoore:
Motivated by monthly check-ins with @mathomH (:heart:), I've begun drafting support for linkml generation into the existing ome-model code generation stack (codename: xsd-fu). The basic scaffolding is now in place, and the template needs to be updated to support more complex types including lists. Once these are properly generated, I'll share a zip of the files so that others can attempt subclassing the types in their own models.

cc: @ome/linkml


joshmoore:
This PR remains a valid example of how to extend the code generation within OME-Model. I've since, however, moved to a simpler solution based on ome-types (see e.g. https://github.com/joshmoore/ome-ld/blob/main/src/mklinkml.py which will likely migrate soon to an official GH organization). Closing for now.

PR #181, opened 2023-11-10
Update license headers for all Java sources of the OME-XML library

sbesson:
While reviewing various boilerplate headers for BSD licensed projects, I noticed that several Java source files under the `ome-xml` component include the terms of the GNU GPL 2 license in their header which is at odds with the component license.

This is most certainly an oversight which can be traced back to the original introduction of the unit concept for the 2015-01 release of the OME Data model in https://github.com/ome/bioformats/pull/1379.

This PR commits the changes after executing `mvn license:update-file-header` under the `ome-xml` component to fix the heders.

The only other notable difference is that the project name is updated from `OME-XML Java library for working with OME-XML metadata structures.` to `OME XML library` as per the POM file. If it is preferable to keep the existing header, happy to change the `name` in `pom.xml`

### DIFF
--- a/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
+++ b/ome-xml/src/main/java/ome/specification/XMLMockObjects.java
@@ -1,6 +1,6 @@
 /*
  * #%L
- * The OME Data Model specification
+ * OME XML library
  * %%
  * Copyright (C) 2003 - 2016 Open Microscopy Environment:
  *   - Board of Regents of the University of Wisconsin-Madison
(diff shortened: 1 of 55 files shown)


joshmoore (approved):
No objections to the updates, though likely a review of the pom.xml values and what's in the header would be worthwhile.

dgault (approved):
License updates all look good from my side.

The header updates for the project name look fine to me also. All the updated files are under the ome-xml folder so using `OME XML library` to match the POM seems correct.

PR #183, opened 2023-12-06
Bump slf4j and logbook versions

dgault:
Bumping versions to match https://github.com/ome/bioformats/pull/3956

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,8 +113,8 @@
          When possible, we advise using the relevant groupId and version
          properties for your dependencies rather than hardcoding them. -->
 
-    <logback.version>1.2.9</logback.version>
-    <slf4j.version>1.7.6</slf4j.version>
+    <logback.version>1.3.13</logback.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -113,8 +113,8 @@
          When possible, we advise using the relevant groupId and version
          properties for your dependencies rather than hardcoding them. -->
 
-    <logback.version>1.2.9</logback.version>
-    <slf4j.version>1.7.6</slf4j.version>
+    <logback.version>1.3.13</logback.version>
```
sbesson, line 116:
```suggestion
    <logback.version>1.3.14</logback.version>
```

sbesson (commented):
Looks like `logback 1.3.14` was released a few days ago  - see https://logback.qos.ch/news.html so proposing we consume this one as there is an associated CVE.
Should I open PRs against `ome-common`, `bioformats`, and `bio-formats-examples` to unify to this version for the upcoming 7.2.0 release?

PR #185, opened 2024-01-30
Add model object API to clear each list of associated objects

melissalinkert:
Suggested in https://github.com/glencoesoftware/ome-omero-roitool/pull/37#pullrequestreview-1850691851

### DIFF
--- a/xsd-fu/templates/java/OMEXMLModelObject.template
+++ b/xsd-fu/templates/java/OMEXMLModelObject.template
@@ -623,6 +623,11 @@ ${customUpdatePropertyContent[prop.name]}
     return new ArrayList<>(${prop.instanceVariableName});
   }
 
+  public void clear${prop.methodName}List()
+  {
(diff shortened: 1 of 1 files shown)


chris-allan (approved):
Looks good to me.

sbesson (approved):
The new proposed API makes sense to me and its naming is consistent with the existing `sizeOf<Object>List()` and `copy<Object>List()` APIs.

Looking at similar examples, https://github.com/ome/bioformats/blob/194a643d02ed766bf57705a3fa3f9e8d1c880cbc/components/bio-formats-tools/src/loci/formats/tools/ImageConverter.java#L599-L604 is probably the most relevant location other than ome-omero-roitool that could be updated to use the new API.

sbesson:
Raised during the weekly formats meeting with @dgault and @jburel, no-one had concerns about including this in the upcoming `ome-model` component release

PR #174, opened 2023-04-25
Remove docs

jburel:
This PR removes the documentation from the repo.
~~Migrated to https://github.com/jburel/ome-model-documentation (for now)~~
Doc now available at https://github.com/ome/ome-model-documentation
RTD https://ome-model.readthedocs.io/en/latest/

It replaces https://github.com/ome/ome-model/pull/147

PR #111, opened 2020-03-31
Remove C++ support

rleigh-codelibre:
To address #104.

- Remove C++ code
- Remove C++ templates
- Remove C++ language support in xsd-fu

I've manually verified that there are no Java code generation differences before and after this change, but you'll want to verify that for yourselves prior to merging.

Will conflict with #108; if you want to merge that first, I'll address the resulting conflicts then.

PR #193, opened 2024-11-20
Build

jburel:
Update actions and Update PyPI publish workflow to use build instead of sdist


sbesson (commented):
While the workflow changes themselves make sense, the Python package is still marked as `experimental` via its name, its content has been unmodified in more than 3 years and I am unaware of any planned work.

This raises the question of whether the Python code under `ome_model` should be extracted into its own repository and maintained separately. This strategy would match what has been done for the C++ implementation (https://github.com/ome/ome-model/pull/111) as well as the Sphinx documentation (https://github.com/ome/ome-model/pull/174).

Listing for discussion at the weekly Formats meeting

Issue #194, opened 2024-11-25
Migrate ome_model Python package to its own repository


sbesson:
The code under https://github.com/ome/ome-model/tree/v6.3.6/ome_model has had little maintenance/development over the last 4 years. 
Functionally, its content is also very different from the rest of this code base which contains:

1- the source XSD files for the OME data model
2- sample XML files for the different schema versions
3- the code-generated Java implementation of the OME model objects

Following the extraction to a separate repository, a decision should likely be made with regard to maintaining it vs archiving it in favor of other community Python projects like https://pypi.org/project/ome-types/. 

_Originally posted by @sbesson in https://github.com/ome/ome-model/pull/193#pullrequestreview-2457871918_

joshmoore:
Especially if we feel comfortable that our existing scripts could be re-written on top of ome-types, I'd be :+1: trying to unify on it.

sbesson:
Agreed, as a proof of concept, I worked on a minimal example upgrading a IDR script generating an HCS companion file using this package to `ome-types`.

The code difference is very reasonable - see https://github.com/IDR/idr0092-ostrop-organoid/pull/2 and the generated OME-XML is consistent with the original one with the only differences being related to attribute ordering and UUIDs.

Based on this minimal investigation, I am further inclined to remove this project and deprecate it in favor of `ome-types`. At the API level, I cannot see anything that I would care about maintaining. If any was raised as valuable, I would propose to discuss how to open it as a contribution against `ome-types` directly. Any valuable experience gained from this project could be also be converted into documentation https://ome-types.readthedocs.io/.

sbesson:
#199 was opened to deprecate the Python code and redirect possible consumers towards using `ome-types`.
The remaining outstanding question is for @tlambert03. Summarizing briefly, https://github.com/ome/ome-model/issues/194#issuecomment-2535308325 describes using `ome-types` directly to generate OME-XML companion files decorating plain TIFF files and turning them into rich OME-TIFF filesets. Would this use case be valuable e.g. as a section in the ome-types documentation?

tlambert03:
Hey @sbesson @joshmoore, thanks for the ping.  Sure, I would happily extend the ome-types documentation.  The use case of turning a regular tiff into an OME-TIFF is a common one that we definitely should have an example of.  Would you have time to contribute that?  (also ok if not)

PR #198, opened 2025-01-04
Apply ruff rules to Python scripts

DimitriPapadopoulos:
I'm assuming all scripts are Python 3 scripts, except this one:
[specification/src/main/resources/transforms/util/write-transforms.py](https://github.com/ome/ome-model/blob/master/specification/src/main/resources/transforms/util/write-transforms.py)

The latter has been marked as a Python 2 script.

### DIFF
--- a/xsd-fu/python/util.py
+++ b/xsd-fu/python/util.py
@@ -27,6 +27,8 @@ import re
 import textwrap
 import time
 
+from pytz import UTC
+
 
 missing = object()
 
@@ -235,8 +237,6 @@ class FixedOffsetTimezone(tzinfo):
     def dst(self, dt):
         return ZERO
 
-
-from pytz import UTC
 
 STDOFFSET = timedelta(seconds=-time.timezone)
 if time.daylight:
--- a/xsd-fu/python/util.py
+++ b/xsd-fu/python/util.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # Copyright (C) 2007 Edgewall Software
 # All rights reserved.
--- a/xsd-fu/python/util.py
+++ b/xsd-fu/python/util.py
@@ -27,7 +27,6 @@ import re
 import textwrap
 import time
 
-from operator import itemgetter
 
 missing = object()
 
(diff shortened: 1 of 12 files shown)


xsd-fu/python/util.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
sbesson, line 1: Same as above, should this be included in #202?

sbesson (commented):
From an initial code review, I don't see concerning changes. Given the modifications to the code generation utilities, we'll need to confirm that the generated classes are unmodified by this change for all supported Python versions.

Several contributions are made against files of a package which has just been deprecated and are scheduled for removal in an upcoming release  (`setup.py`, `ome_model/experimental.py`, `test/unit/*`)  - see #194 for more details. Given the scope of this work, we might pioritize the clean up of these classes first and update you when this is completed and conflicts should be resolved so that only changes against `xsd-fu` utility are evaluated

xsd-fu/python/util.py
DimitriPapadopoulos: I cherry-picked the commit from #202.

sbesson:
@DimitriPapadopoulos with #206 merged, can you fix the conflicts on this PR by merging `origin/master` into your branch?

DimitriPapadopoulos:
Done

sbesson:
Thanks. A diff of the Java classes generated by the `xsd-fu` utility against the source files generated by the current HEAD of this repository revealed no functional change introduced by this PR. I am labelling it for inclusion in the nightly CI builds so that we can test all integration tests keep passing. We can then apply a formal round of code review.

sbesson (approved):
I generated and compared the output of `mvn clean generated-sources` executed using all supported Python versions (3.9, 3.10, 3.11, 3.12) against the same files generated from the current HEAD of this repository using Python 3.12.
In all cases, `diff -r` showed no difference and `diff -rs` confirmed the files were identical.

Most of the changes are related to code formatting and/or also modernization of some of the Python code from this utility. I'll leave @melissalinkert to take a look for anything of potential concern. Overall no objection to getting this in.

melissalinkert (approved):

PR #200, opened 2025-01-07
Delete dead code

### LINKED ISSUE #197: `self.name == 'Annotation'`
This expression is useless as is:
https://github.com/ome/ome-model/blob/0df23970b8c73bafbbf1b3ea90dbfca44c406e1d/xsd-fu/python/ome/modeltools/object.py#L61-L62

I think the equality operator `==` in
```python
     or self.name == 'Annotation' 
```
should be changed to an assignment `=`:
```python
     or self.name = 'Annotation'
```

DimitriPapadopoulos:
Fixes #197.

### DIFF
--- a/xsd-fu/python/ome/modeltools/object.py
+++ b/xsd-fu/python/ome/modeltools/object.py
@@ -58,8 +58,6 @@ class OMEModelObject(OMEModelEntity):
         self.isUnique = False
         self.isImmutable = False
         self._isGlobal = False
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Nightly CI builds were green with this included. Also tested locally by running the maven build with and without this PR and comparing the output of `ome-xml/target/generated-sources`. The files were binary identical and confirm this code was unused

PR #201, opened 2025-01-08
Add a trailing newline for POSIX compliance

DimitriPapadopoulos:
Most files end with a newline and are POSIX compliant. Add a final newline to the few files that lack it.


sbesson (approved):
Will wait for a successful nightly build to complete but this looks pretty straightforward

PR #202, opened 2025-01-08
Use the default encoding of Python files

DimitriPapadopoulos:
The default encoding of Python 3 files is UTF-8, there is no need to specify it.

These are either plain ASCII files or already UTF-8 files.


sbesson (approved):

DimitriPapadopoulos:
Moved to #198.

Issue #207, opened 2025-01-22
`specification/src/main/resources/transforms/util/write-transforms.py` is still a Python 2 script


DimitriPapadopoulos:
[`resources/transforms/util/write-transforms.py`](https://github.com/ome/ome-model/blob/master/specification/src/main/resources/transforms/util/write-transforms.py) is still a Python 2 script. For example, `print` is not a function, it is used without parentheses.

Is this script still needed? If so, shouldn't it be converted it to Python 3?

EDIT: There may be multiple scripts in this situation:
* `specification/src/main/resources/transforms/util/write-transforms.py`
* `xsd-fu/python/generateDS/generateDS.py`
* `xsd-fu/python/generateDS/process_includes.py`

sbesson:
Re `write-transforms.py`, I believe this has been used to generate https://github.com/ome/ome-model/blob/a3c7457b5ae28ecfd1ca42612435cd68b1f0768a/specification/src/main/resources/transforms/ome-transforms.xml#L123 i.e. the XML file specifying the transformation path between versions of the model.

This is used as a one-off rather than as part of the build system. Since there hasn't been a specification release since 2016-06, there has been no real driver to upgrade this script. Upgrading should be fairly straightforward though and we can take advantage of the XML file to ensure it produces the expected outcome.

I am pretty sure the `generateDS` module is used e.g. in https://github.com/ome/ome-model/blob/a3c7457b5ae28ecfd1ca42612435cd68b1f0768a/xsd-fu/python/ome/modeltools/generateds.py#L33

DimitriPapadopoulos:
About `xsd-fu/python/generateDS/generateDS.py`, the reason I was asking is that it contains a `has_key` call from Python 2, not available any more in Python 3:
https://github.com/ome/ome-model/blob/f6a81fc40924379f0607e0ca1e0a886c3752c224/xsd-fu/python/generateDS/generateDS.py#L907-L908

I'll fix that in a different PR.

PR #210, opened 2025-01-22
Apply ruff preview rules to Python scripts

DimitriPapadopoulos:
Are all scripts supposed to have switched to Python 3 (except for those in #207)?


sbesson (approved):
Similar to the testing of https://github.com/ome/ome-model/pull/198, I did not find any difference in the code-generated sources with and without this PR. CI tests are green

PR #211, opened 2025-01-23
Port script to Python 3

### LINKED ISSUE #207: `specification/src/main/resources/transforms/util/write-transforms.py` is still a Python 2 script

DimitriPapadopoulos:
I have checked that the script regenerates an identical `specification/src/main/resources/transforms/ome-transforms.xml` file.

Fixes #207.

### DIFF
--- a/specification/src/main/resources/transforms/util/write-transforms.py
+++ b/specification/src/main/resources/transforms/util/write-transforms.py
@@ -82,7 +82,7 @@ def shortest_path(from_schema, to_schema, min_quality):
 
 # the style of transform file names
 
-name_pattern = compile('^(.+)\-to\-(.+)\.xsl$')
+name_pattern = compile(r'^(.+)\-to\-(.+)\.xsl$')
 
 
 # scan the current directory to determine the schemas and transforms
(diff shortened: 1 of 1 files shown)


sbesson (approved):
```
(venv3.12) sbesson@Sebastiens-MacBook-Pro-3 transforms % python util/write-transforms.py > ome-transforms.xml               
(venv3.12) sbesson@Sebastiens-MacBook-Pro-3 transforms % git diff
(venv3.12) sbesson@Sebastiens-MacBook-Pro-3 transforms % 
```

Thank you @DimitriPapadopoulos

PR #217, opened 2025-02-05
Apply `ruff format` to Python code

DimitriPapadopoulos:


sbesson:
Thanks for the contribution. However after discussion with the @ome/formats team, the general consensus is in addition to being a significant cost in terms of reviewing this PR which modifies thousands of lines of code, the value of enforcing a Python formatting style is not worth it at this stage.

DimitriPapadopoulos:
I understand. The idea was to **not** to review the PR. Instead, a maintainer could have rerun `ruff format` locally and verified that the generated code is identical. While not all `ruff check` fixes are safe, `ruff format` is supposed to be totally safe - it merely reformats code (spaces, line breaks, etc.) without any substantial modification.

PR #222, opened 2025-06-24
Update publication to use Central Portal

sbesson:
See https://central.sonatype.org/news/20250326_ossrh_sunset/ for more context as well as https://github.com/ome/ome-common-java/pull/101 and https://github.com/ome/ome-contributing/pull/37

This primary adjusts the `release` profile used during the `deploy` phase to use `central-publishing-maven-plugin` instead of `nexus-staging-maven-plugin`.
The `maven-compiler-plugin` plugin is also upgraded to 3.14.0 and the release target set to 8. This should ensure release artifacts built with either version of the JDK are binary compatible with Java 8 - see https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html#usage-on-jdk-8 for more information
Other changes are primarily cosmetic - updating a few outdated sections and replacing HTTP by HTTPS

### DIFF
--- a/ome-xml/pom.xml
+++ b/ome-xml/pom.xml
@@ -16,11 +16,11 @@
 
   <name>OME XML library</name>
   <description>A library for working with OME-XML metadata structures.</description>
-  <url>http://www.openmicroscopy.org/site/support/ome-model/ome-xml/java-library.html</url>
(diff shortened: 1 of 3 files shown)


joshmoore (approved):
:+1:

PR #230, opened 2026-05-13
Require Java 11 at runtime, build on Java 21/25

melissalinkert:
See https://www.openmicroscopy.org/2026/03/24/end-of-java-8-support.html. Similar to https://github.com/ome/ome-metakit/pull/33.

The build matrix should probably be pared down again, but wasn't 100% sure what we want at this point so open to discussion.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,7 +15,26 @@ jobs:
       matrix:
         java: [11, 17, 21, 25]
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
+        exclude:
+          - java: 11
+            python: 3.11
+          - java: 11
+            python: 3.12
+          - java: 11
+            python: 3.13
+          - java: 17
+            python: 3.10
+          - java: 17
+            python: 3.12
+          - java: 17
+            python: 3.13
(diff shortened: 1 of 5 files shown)


.github/workflows/maven.yml
```diff
@@ -13,55 +13,24 @@ jobs:
       contents: read
     strategy:
       matrix:
-        java: [8, 11, 17, 21]
+        java: [11, 17, 21, 25]
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
-        exclude:
```
sbesson, line 19: A consequence of this removal is the explosion of the number of jobs. Proposing to drop Python 3.9 from the list above and only cover all the Python versions for a particular Java version e.g. 17 or 21.

sbesson (changes requested):

.github/workflows/maven.yml
melissalinkert: See 1c544cd5. That should test one Python version each for Java 11/17/25, and all Python versions for Java 21. The matrix prior to this PR ended up excluding everything except Java 11 across all Python versions, which I think is not what we actually want.

sbesson (approved):

sbesson (approved):
Merging and releasing as per the passing nightly builds

sbesson:
This has been released to Maven Central as `org.openmicroscopy:ome-xml:6.6.0`

PR #234, opened 2026-08-04
Python cleanup

Darlokt:
Hej,

While investigating the CI failure of ome/ome-model#233, and tracking them down to the larger filter list enabled by `ruff` `0.16.0` (see https://astral.sh/blog/ruff-v0.16.0), I found broader problems with the tooling, including that the repository contained a mixture of Python 2- and Python 3-era code, vendored `generateDS` sources, legacy entry points, duplicated configuration, no version pinning, etc.

Rather than adding increasingly broad Ruff exceptions around code that was difficult to maintain, I found that a clean rewrite around `xmlschema`, replaces the custom schema-processing implementation, providing a clearer and more maintainable foundation for validating and traversing OME XML Schema documents.

## Changes

Replace the legacy `generateDS`-based implementation and `xsd-fu/xsd-fu` script with an installable `xsd-fu` package under `xsd-fu/src`.

The new implementation adds:

- XML Schema loading and validation through `xmlschema`
- A normalized schema graph for elements, attributes, simple types, choices, groups, namespaces, and dependencies
- A typed OME model layer for objects, properties, enumerations, references, units, substitution groups, annotations, and metadata relationships
- Java configuration for type mappings, packages, naming, units, inheritance, templates, and generated source paths
- Genshi-based rendering of Java model and metadata sources
- Explicit generation settings and model-processing errors
- A package entry point through both `xsd-fu` and `python -m xsd_fu`

The command-line interface now supports:

- Generating model sources, metadata sources, or both
- Dry-run generation
- Listing generated files
- Listing schema and template dependencies
- Custom template directories
- Java package overrides
- Validation of `MetadataStore`, `MetadataRetrieve`, and `MetadataConverter` APIs

Generation is performed in memory before writing any files. Output paths are checked for conflicts, generated files are ordered deterministically, and writes use temporary files followed by atomic replacement.

Move all Java templates into the package so installed wheels can generate sources without depending on the repository layout.

Remove the legacy Python package tree, vendored `generateDS` implementation, obsolete utility code, and old executable. The generated output remains compatible with the existing OME Java APIs, as verified against the frozen output contract.

Update `ome-xml/pom.xml` to invoke the locked `xsd-fu generate` command once during `generate-sources`, producing both model and metadata APIs in a single validated pass.

## Python tooling

Modernize the project configuration by:

- Renaming the distribution to `xsd-fu`
- Requiring Python 3.13
- Adding `xmlschema`
- Using `uv_build` for packaging
- Publishing the `xsd-fu` console script
[...]

PR #235, opened 2026-08-04
Add dependabot

Darlokt:
Hej,

As mentioned in ome/ome-model#234 this adds a minmal dependabot config, to automatically open pull requests for up to date dependencies/bumps.

I for now set it up to cover the Python dependencies, (after ome/ome-model#234 is merged) through `uv`, `maven` and the GitHub Actions versions.

With ome/ome-model#234 currently outstanding updates are:
Dependency | Outstanding updates
-- | --
GitHub Actions | 0
`uv` | 0
`maven` | 25

With the `maven` updates applied, everything builds correctly and passes the test suite. But as the `maven` updates are quite big jumps, it should be tested more thoroughly with downstream.


melissalinkert:
We agree that establishing a pattern for keeping Maven/plugins up to date is useful (see https://github.com/ome/bioformats/issues/4469), but we prefer to coordinate this within the wider ecosystem of OME repositories instead of adding dependabot to ome-model right now.

